PR #4, opened 2017-11-18
Removing fake-hwclock is not a must.

marcin-sielski:


makermelissa:
Closing because this is so old and much has changed since then.

PR #11, opened 2018-04-08
added uninstall_fbcp call when answering 'no' to hdmi mirroring

Lightfire228:


Lightfire228:
An explanation of the changes might be a good idea.
I need the pitft to display xserver, but during testing, I like to have tty1-6 on the HDMI so I can see what I'm doing.  So I give the installer script the inputs "3, 1, n, n, n", which sets it for 2.8c, at landscape, no console, no hdmi mirroring, no reboot (I do that later).  
I also include an fbdev conf file at `/usr/share/X11/xorg.conf.d/99-fbdev.conf`

```
Section "Device" 
  Identifier "myfb"
  Driver "fbdev"
  Option "fbdev" "/dev/fb1"
EndSection
```

I was playing around with the installer script and noticed that if you answered yes to hdmi screen mirroring, the effect remains when you re-run the script and say no to hdmi screen mirroring.  The first commit adds a call to the unused function `uninstall_fbcp` when answering no to hdmi screen mirroring.

Then I noticed that the pointer for 2.8c screens had the wrong rotation settings.  I tracked that down to this if 
```
    if [ -e /etc/lightdm ]; then
        info PITFT "Updating X11 default calibration..."
        update_xorg || bail "Unable to update calibration"
    fi
```
being nested in the hdmi screen mirroring if.  So the second commit moves the code above to run regardless of hdmi mirroring

Lightfire228:
closing since it has been a year, and my PR is probably outdated now

PR #12, opened 2018-04-14
Set C99 for Raspbian jessie

chris-cartwright:


ladyada:
thanks! some retropie-folks needed this!

PR #18, opened 2018-04-30
Keep cron package

kohlerdominik:
Is there any reason, why cron should be removed? 
I think, for a lot of application it's still a usefull function.

Of course, we should keep fake-hwclock as well, like #4 suggests. Or, if there are any reasons to remove hwclock, then maybe we can install "ntp" instead?


kohlerdominik:
fake-hwclock doesn't work as expected. We should install ntp instead.

Issue #5, opened 2017-12-04
NTP Time ?


JacoFourie:
After running the script the Pi will default to 3 November 2016. NTP time is also missing. So you dont have accurate date and time.

jacobalberty:
It seems to be a dhcpcd issue, dhcpcd seems to be trying to write something before calling its exit scripts, /run/systemd/timesyncd.conf.d/ likely needs to be mounted as a tmpfs to fix /etc/dhcp/dhclient-exit-hooks.d/timesync but something else is happening to keep that from even running.

jacobalberty:
Well, I guess maybe thats not the right hook? I'm confused there seems to be multiple hooks /lib/dhcpcd/dhcpcd-hooks/50-ntp.conf may be the right one to worry about and it does seem to be running but not working probably need at least /var/run/ /var/run/dhcpcd/ntp.conf to be on tmpfs, but so fat at least that isnt enough

pl31:
May it is this one:

> Note: The service writes to a local file /var/lib/systemd/timesync/clock with every synchronization. This location is hard-coded and cannot be changed. This may be problematic for running off read-only root partition or trying to minimize writes to an SD card.

From here:
https://wiki.archlinux.org/index.php/systemd-timesyncd

pl31:
After running the script and booting:

```
pi@raspberrypi:~ $ timedatectl status
      Local time: Thu 2016-11-03 18:18:49 CET
  Universal time: Thu 2016-11-03 17:18:49 UTC
        RTC time: n/a
       Time zone: Europe/Berlin (CET, +0100)
 Network time on: yes
NTP synchronized: no
 RTC in local TZ: no
```

Trying to reconfigure:
```
pi@raspberrypi:~ $ sudo timedatectl set-ntp 0

Failed to set ntp: File /etc/systemd/system/sysinit.target.wants/systemd-timesyncd.service: Read-only file system
```

... but no clue how to fix his bug or if it is even related.

pl31:
Seems not to be an readonly issue:

```
pi@raspberrypi:~ $ sudo mount -o remount,rw / 
pi@raspberrypi:~ $ sudo service systemd-timesyncd restart
Warning: systemd-timesyncd.service changed on disk. Run 'systemctl daemon-reload' to reload units.
pi@raspberrypi:~ $ sudo systemctl daemon-reload
pi@raspberrypi:~ $ sudo service systemd-timesyncd restart
[...]

PR #22, opened 2018-05-12
Keep cron & install ntp

kohlerdominik:
Keep cron so time tasks are still possible. If there's no internet connection, of course only interval will be correct, but not the execution time.
If there's an internet connection, ntp will set the time after boot, and then even the execution time will be right.


mamoit:
IMHO the best compromise to solve #5.
@kohlerdominik thanks for the PR

kohlerdominik:
Any change this get merged in @PaintYourDragon ?

PR #28, opened 2018-06-22
add service unit for managing fbcp under systemd

brennen:
Checks if we're under systemd, and if so, ensures that fbcp isn't mentioned in /etc/rc.local, and installs a new unit file to start fbcp at boot, after network.target.

If not under systemd, retains the old rc.local behavior.

I note this is the only place that the $SYSTEMD var actually seems to get checked.  I can't, at this time, vouch for what would happen on a sysvinit box.


ladyada:
looks good!

PR #30, opened 2018-10-08
Import occi util from Adafruit-Occidentalis repo

brennen:
This adds the `occi` config helper, now with an `--install` option to install system-wide and start on boot via systemd.


ladyada:
perl!

brennen:
:camel:

PR #37, opened 2018-11-13
add @ladyada's libgpiod install instrux to script

brennen:
See also: https://github.com/adafruit/Adafruit_Blinka/issues/43

### DIFF
--- a/libgpiod.sh
+++ b/libgpiod.sh
@@ -1,0 +1,44 @@
+#!/bin/sh
+
+# Instructions!
+# cd ~
+# wget https://raw.githubusercontent.com/adafruit/Raspberry-Pi-Installer-Scripts/master/libgpiod.sh
+# chmod +x libgpiod.sh
+# ./adafruit-pitft.sh
+
+
+echo "Installing build requirements - this may take a few minutes!"
+echo
+
+sudo apt-get update && sudo apt-get install -y \
+   autoconf \
+   autoconf-archive \
+   automake \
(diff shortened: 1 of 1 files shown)


libgpiod.sh
```diff
@@ -0,0 +1,44 @@
...
+echo
+
+cd "$build_dir"
+git clone git://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git .
+
+echo "Building libgpiod"
+echo
+
+export PYTHON_VERSION=3
+./autogen.sh --enable-tools=yes --prefix=/usr/local/ --enable-bindings-python CFLAGS="-I/usr/include/python3.5/" \
```
ladyada: is there a way to get the current py3 include dir instead of hardcoding?

libgpiod.sh
```diff
@@ -0,0 +1,46 @@
...
+echo "Cloning libgpiod repository to $build_dir"
+echo
+
+cd "$build_dir"
+git clone git://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git .
+
+echo "Building libgpiod"
+echo
+
+include_path=`python3 -c "from sysconfig import get_paths; print(get_paths()['include'])"`
```
brennen: This seems to work.

libgpiod.sh
brennen: See latest commit with `sysconfig`.

PR #38, opened 2018-11-15
add some hacky file copies for libgpiod python bindings

brennen:
I'll suss out the correct way to do this at some point; it works for the moment.


ladyada:
oi ok for now :D

PR #39, opened 2018-11-30
pitft: add ExecStartPre sleep of 10 seconds before starting fbcp

brennen:
Hopefully fixes glitchiness described in [this forum thread](https://forums.adafruit.com/viewtopic.php?f=47&t=142746&p=707588).

Tested on Pi 1; appears to be working.


ladyada:
tested!

PR #43, opened 2018-12-20
Add -f S16 option

donkeykey:
It's better to add -f S16 option into aplay command so that improve muffled sound.

### DIFF
--- a/i2samp.sh
+++ b/i2samp.sh
@@ -445,7 +445,7 @@ EOL
 Description=Invoke aplay from /dev/zero at system start.
 
 [Service]
-ExecStart=/usr/bin/aplay /dev/zero -f S16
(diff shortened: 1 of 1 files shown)


ladyada:
ooh can you try
`ExecStart=/usr/bin/aplay -D default -t raw -r 48000 -c 2 -f S16_LE /dev/zero`
does that sound even better?

donkeykey:
OK, I tried it with this command. 
```
/usr/bin/aplay -D default -t raw -r 48000 -c 2 -f S16_LE /dev/zero
```
It sounded equally good!

ladyada:
thanks key! can you update your PR for that longer line as some people said it worked best for them :)

PR #51, opened 2019-05-28
aplay: change rate to 44100 to avoid unnecessary CPU usage

nicokaiser:
The aplay service plays `/dev/zero` to avoid clicks. The current version uses a sampling rate of 48 kHz, while ALSA is configured to use 44.1 kHz  (some lines above in `i2samp.sh`).
This leads to ALSA/dmixer resampling the silence to the right rate, which is unnecessary. CPU usage is much better if the sampling rate is set correctly in aplay (tested on a Raspberry Pi 3).

### DIFF
--- a/i2samp.sh
+++ b/i2samp.sh
@@ -445,7 +445,7 @@ EOL
 Description=Invoke aplay from /dev/zero at system start.
 
 [Service]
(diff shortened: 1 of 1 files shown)


ladyada:
merged, we'll see if anyone complains :D

ladyada:
thanx!

PR #58, opened 2019-07-16
Update library reference to support Pi 4

wibbleboy:
The existing commit reference for hzeller's library points to a version of which doesn't support the recently released Raspberry Pi 4. Edited the script to point to the new commit before running it and it started working fine on the Pi4.

### DIFF
--- a/rgb-matrix.sh
+++ b/rgb-matrix.sh
@@ -7,7 +7,7 @@
 # we reference a specific commit (update this as needed):
 GITUSER=https://github.com/hzeller
 REPO=rpi-rgb-led-matrix
-COMMIT=814b79b5696d32dd1140304b41a1ec0068bb271a
+COMMIT=5549b6a8539f6191b0042a9882844a36716acebd
 
 if [ $(id -u) -ne 0 ]; then
 	echo "Installer must be run as root."
--- a/rgb-matrix.sh
+++ b/rgb-matrix.sh
@@ -7,7 +7,7 @@
 # we reference a specific commit (update this as needed):
 GITUSER=https://github.com/hzeller
 REPO=rpi-rgb-led-matrix
-COMMIT=5549b6a8539f6191b0042a9882844a36716acebd
+COMMIT=e3dd56dcc0408862f39cccc47c1d9dea1b0fb2d2 
 
 if [ $(id -u) -ne 0 ]; then
 	echo "Installer must be run as root."


rgb-matrix.sh
```diff
@@ -7,7 +7,7 @@
 # we reference a specific commit (update this as needed):
 GITUSER=https://github.com/hzeller
 REPO=rpi-rgb-led-matrix
-COMMIT=814b79b5696d32dd1140304b41a1ec0068bb271a
+COMMIT=5549b6a8539f6191b0042a9882844a36716acebd
```
hzeller: Can you change this to the next commit e3dd56dcc0408862f39cccc47c1d9dea1b0fb2d2 ?
I've added more messaging around the fact that the new version also can run as non-root, but better point out the limitations. This will safe me many support emails :D

wibbleboy: Updated commit as requested.

wibbleboy:
> thanks, have you tried the tutorial to verify all the demos/commands still work? the API has changed before, just want to make sure if it did again

Yes, I checked that each of the demos were working fine on the Pi4, but haven't been able to check that nothing is broken on previous versions of the Pi.

ladyada:
@PaintYourDragon fyi, im willing to take the risk that it works on previous boards too

ladyada:
@wibbleboy thanks!

hzeller:
pi matrix author here: I didn't change the API noticeably since last time. I mostly changed the detection so that Pi4 and Pi3a+ is properly detected.

So everything should be working fine including Pi4.

Something that might come up and might be useful in a tutorial: since the Pi4 can be much faster outputting stuff to GPIO (I measured in the order of 130Mhz, almost twice as fast as Pi3),  the --led-slowdown-gpio might need to be set to 2 for longer chains or slower panels.

wibbleboy:
> Something that might come up and might be useful in a tutorial: since the Pi4 can be much faster outputting stuff to GPIO (I measured in the order of 130Mhz, almost twice as fast as Pi3), the --led-slowdown-gpio might need to be set to 2 for longer chains or slower panels.

I'll be testing against 4 64x64 panels (not sure if this could be considered a "long chain" ;), but I'll feed back if I see anything out of the ordinary.

ladyada:
@hzeller thanks for your work updating and maintaining this library ! i've updated the guide to indicate Pi 4's should start with a slowdown of 2 :)

PaintYourDragon:
Tested on Pi 4, 3 and 1B+, works in all cases. **Thanks everyone!**

Pi 4 required --led-slowdown-gpio=4 …at least with the particular matrix I was testing and with no GPIO4-18 jumper. Added notes in a couple places in guide to try values >2 if needed.

Everything’s good with the latest Raspbian Buster (July 10th freshness date) … was previously using the initial June release, that one had some issues with the matrix install script. That release was very short-lived and isn’t likely to affect very many. If anyone encounters installation trouble, this might be the first thing to check.

hzeller:
Ok, sounds like we need to allow more slowdown for Pi4s.
I've extended the allowed range for this value to 5 in commit d73456c2164b182d06c9c87513fe3913f87a1994 ( https://github.com/hzeller/rpi-rgb-led-matrix/commit/d73456c2164b182d06c9c87513fe3913f87a1994 )

Issue #74, opened 2020-02-24
CMAKE argument deprecated in Pitft.sh


ankonaskiff17:
PITFT.sh
I am new at this but when I run the PiTFT.sh easy install after selecting the screen size and orientation it starts trying to install various libraries but it gets to a step:

 [PITFT] Adding FBCP support...
Installing cmake...
W: --force-yes is deprecated, use one of the options starting with --allow instead.
Downloading rpi-fbcp...
Uncompressing rpi-fbcp...

Something is not right with the script and among other things you can manually create the /dev/input/touchscreen directory and on the reboot that is the final step, it deletes the touchscreen directory. /dev/input remains but touchscreen is removed,

ladyada:
deprecation warning fixed

PR #79, opened 2020-04-01
replace deprecated --force-yes

OopsDoh:
See [https://manpages.debian.org/stretch/apt/apt-get.8.en.html](https://manpages.debian.org/stretch/apt/apt-get.8.en.html)

**--force-yes**
Force yes; this is a dangerous option that will cause apt to continue without prompting if it is doing something potentially harmful. It should not be used except in very special situations. Using force-yes can potentially destroy your system! Configuration Item: APT::Get::force-yes. This is deprecated and replaced by --allow-downgrades, --allow-remove-essential, --allow-change-held-packages in 1.1.

### DIFF
--- a/adafruit-pitft.sh
+++ b/adafruit-pitft.sh
@@ -345,7 +345,7 @@ function uninstall_console() {
 
 function install_fbcp() {
     echo "Installing cmake..."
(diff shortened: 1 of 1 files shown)


OopsDoh:
Related issue: #74

PR #81, opened 2020-04-10
Add option to create X11 xorg.config.d configs when fbcp and/or lightdm not installed

thess:
After installing xorg without any desktop manager, the install script fails to configure X11 for the TFT framebuffer device /dev/fb1. It also dost not add the touchscreen calibration info. This patch adds an option to update X11 when neither the console or fbcp are selected. 

Creates: **05-pitft.conf** for /dev/fb1 Device and **20-calibration.conf** with correct touchscreen rotation and transform params.


makermelissa:
Thanks, we redid install script and touch rotation is currently working correctly.

Issue #85, opened 2020-04-22
PiTFT: Uninstall Option


makermelissa:
In order to make it easier for users to remove the kernel level driver, it would be great to have an uninstall option that went through and cleaned up everything that it installed so users could use it with Python.

Trudko:
@makermelissa Do you please have an idea, what would be workaround till then? I've installed kernel driver but would love to use it with python now.

makermelissa:
Hi. Look in /boot/config.txt. You’ll need to remove the appropriate pitft device tree overlay for now.

makermelissa:
Ok, I had to do this manually on one of my Raspberry Pis today and I had to remove some lines from `/etc/modules` as well.

Trudko:
@makermelissa thank you, could you maybe go more into details what exactly have you changed?

makermelissa:
Fixed via #90

makermelissa:
It's available, so you can look at the script now @Trudko  ;)

PR #90, opened 2020-05-16
Added uninstall option to PiTFT

makermelissa:
I added an uninstall option to help folks who want to easily use RGB_display (or the upcoming displayio) on their Pi. It makes use of existing uninstall functions and breaks out some that were part of an install into their own functions.


makermelissa:
Fixes #85

PR #94, opened 2020-06-07
Ntp fix etc

SkyperTHC:
1. tmpfs and systemctl on latest raspberry os breask ntp because /tmp is private to ntp. This results in /etc/resolv.conf symbolic link to became unavailable to ntpd. 

2. read-only fs should never be fsck'ed if 'reliability' and 'boot at all cost even fs is corrupted' is the goal. Otherwise linux will check fs every 6 month (including read only fs) and if corruption is found will wait for user to confirm to continue - you dont want this.

3. fixed vim complaining about ~/.viminfo not being writable.


kohlerdominik:
@PaintYourDragon could you review this?

PaintYourDragon (approved):
I’ll merge it, but FYI this script is no longer used in the corresponding guide, which now just focuses on the read-only /boot feature that’s part of Raspberry Pi OS

Issue #102, opened 2020-07-27
Installation of st7789 display via adafruit-pitft.sh fails


OopsDoh:
I recently updated RaspberryPi OS (and tried a new installation as well) and now the installation of the [2.0'' 320x240 spi display](https://www.adafruit.com/product/4311) (st7789) with [adafruit-pitft.sh](https://github.com/adafruit/Raspberry-Pi-Installer-Scripts/blob/master/adafruit-pitft.sh) fails (installation support [requested/suggested here](https://github.com/adafruit/Raspberry-Pi-Installer-Scripts/issues/80) and added in [this PR](https://github.com/adafruit/Raspberry-Pi-Installer-Scripts/pull/86) ).

I think the reason is that [fbtft_device and flexfb are gone in 5.4](https://github.com/notro/fbtft/wiki#fbtft_device-and-flexfb-are-gone-in-54)

In previous installations the st7789 displays have been enabled by adding fbtft_device init-sequence in /boot/config.txt

Unfortunaltely this method doesn't work anymore with a current RaspberryPi OS with Linux 5.4 kernel.

For testing I also tried to select the 2.2'' display in the installation script: The 2.0'' display is functioning but only shows an distorted image. No wonder because the 2.2 display is no st7789 driver. But when selecting this wrong display size, there is an dtoverlay added in the boot/config.txt

I guess there has to be added an [Device Tree overlay](https://github.com/notro/fbtft/wiki/FBTFT-RPI-overlays) for the st7789 displays. Unfortunately I didn't yet understand how this works.

[...]

PR #103, opened 2020-07-27
St7789 redux

ladyada:
without fbflex, we need a kernel module for any st7789 device!


ladyada:
if you could try this on both the lil 1.13" minipitft and also the 1.3" or 1.5" TFT bonnet/miniPiTFT - it should work on the very latest pios

makermelissa:
I'm not getting anything on the 1.3" TFT Bonnet. Is the driver turning on the backlight?

makermelissa:
Yeah, I turned it on via Blinka and I see the display now.

makermelissa:
Works great on 1.14"

makermelissa:
After looking at the code, the issue appears to be that the 1.3" mini PiTFT uses Pin 22 for the Backlight and the 1.3" Bonnet uses Pin 26 for the backlight. Can we set both pins to high so that it will work on either one?

ladyada:
ok i changed the DTS, please try re-installing

makermelissa:
Ok, will do. :)

makermelissa:
I can't seem to get it working. I updated the branch, removed any generated files from a previous run and am getting the following output:
```
[PITFT] Updating /boot/config.txt...
/boot/overlays/drm-minipitft13.dtbo: Warning (unit_address_vs_reg): /fragment@0/__overlay__/spidev@0: node has a unit name, but no reg property
############# UPGRADING KERNEL ###############
Hit:1 http://archive.raspberrypi.org/debian buster InRelease
Hit:2 http://raspbian.raspberrypi.org/raspbian buster InRelease
Reading package lists... Done
Building dependency tree
Reading state information... Done
All packages are up to date.
Reading package lists... Done
Building dependency tree
Reading state information... Done
Calculating upgrade... Done
The following package was automatically installed and is no longer required:
  rpi-eeprom-images
Use 'sudo apt autoremove' to remove it.
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
WARNING : Kernel was updated, please reboot now and re-run script!
```

ladyada:
hmm ok ill get one from the shop - i dont seem to have one here to test with, ill look again

makermelissa:
FWIW, I tried this fresh install again because I messed something up when trying to reset it and it installs fine, it just doesn't display anything on the bonnet. The 1.3" Mini PiTFT tested fine with the latest changes.

ladyada:
@makermelissa try now. its hacky but should light up the backlite on the bonnet :P

makermelissa:
One other question. Since the driver folder is separate of the install script, do we need to update instructions in the guides or does/can the installer script wget the driver folder from here if it doesn't already exist?

makermelissa:
I only tested console, so I am going to test the FBCP method as well based on comments in #102.

makermelissa:
I haven't been able to get the FBCP driver to work at all. It's a black screen with the ST7789 (both the 2.0" and 1.3" displays) and it looks odd on the ILI9341:
3B99132B-6575-4E53-8F14-434A1A07B563_1_105_c

ladyada:
fbcp doesnt work with even the older displays anymore?

makermelissa:
Not on the Pi 4 (with 2 HDMI ports). I can see if it works any differently on the Pi 3 with a single HDMI.

ladyada:
ya dya mind? id like to narrow down if its an issue with the pi4, 5.4 or a combo of the two!

makermelissa:
Nope, I don't mind. :)

ladyada:
thanx - particularly on the ili934x - its disconcerting cause it used to work!

makermelissa:
Ok, ILI9341 looks fine on a Pi 3. I'll try the 2.0" display.

makermelissa:
2.0" display is not working at all now, even in console mode. It's like just installing FBCP breaks something. I tried it both with and without HDMI plugged in.

makermelissa:
1.3" isn't working either anymore. Come to think of it, I had a lite image that I tried the console version on, so the different version may also be a factor.

ladyada:
ok well given this at least fixes console mode, it is merged

PR #104, opened 2020-08-02
5.4 fixfbcp

ladyada:
@makermelissa this should at least fix fbcp mirroring for old displays (2.2", 2.8", 3.2" and 3.5")


makermelissa:
So far everything is looking really good including it working on the 2.0" ST7789. 
IMG_2334

ladyada:
yay ok merging and we'll see what else bubbles up!

makermelissa:
Sounds good

PR #110, opened 2020-08-12
Added legacy option for systems with older linux headers

### LINKED ISSUE #109: libgpiod.sh fails on older systems when Linux Headers are <= 5.5.0
These are probably the systems that need this script the most. You can check what the latest Linux Headers available for a system are by typing:
`sudo apt-cache show linux-headers-generic | grep Version`

A workaround that I discovered is by checking out an earlier branch of libgpiod. I used 1.4.2 and was successfully able to get it to work with the latest version of Blinka.

However, I propose adding a flag like --legacy and just checking out whatever legacy version we hardcode into the script. So for instance, if we go with 1.4.2, then right after `git clone git://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git .`, we can check if a legacy flag is set and just call:
```
git checkout v1.4.2 -b v1.4.2
```
After that, the rest of the script should work just fine.

makermelissa:
Fixes #109. Tested both paths on a system that experienced this issue and they behaved as expected.

### DIFF
--- a/libgpiod.sh
+++ b/libgpiod.sh
@@ -6,6 +6,21 @@
 # chmod +x libgpiod.sh
 # ./libgpiod.sh
 
+is_legacy=0
+
+# Loop through arguments and process them
+for arg in "$@"
+do
+    case $arg in
+        -l|--legacy)
+        is_legacy=1
+        shift
+        ;;
+        *)
+        shift
+        ;;
+    esac
+done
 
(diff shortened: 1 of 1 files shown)


makermelissa:
Ah, I thought you meant just the PiTFT script. Yeah, this would be a good one too. :)

PR #111, opened 2020-08-14
Converted adafruit_fanservice to python

makermelissa:


adafruit_fanservice.py
```diff
@@ -0,0 +1,131 @@
...
+    if r == 0:
+        return True
+    else:
+        if not suppress_message:
+            ColorPrint.print_fail(err.decode("utf-8"))
+        return False
+
+def main():
+    os.system('clear')
+    print("This script will install Adafruit")
```
ladyada, line 64: use """ to make one block https://stackoverflow.com/questions/34980251/how-to-print-multiple-lines-of-text-with-python

adafruit_fanservice.py
```diff
@@ -0,0 +1,131 @@
+import sys
+import os
+import subprocess
+
+group = 'ADAFRUIT'
+
+class ColorPrint:
```
ladyada, line 7: try using clint instead
https://github.com/adafruit/Adafruit-Metro-328-Tester/blob/master/megaprog.py#L6

adafruit_fanservice.py
```diff
@@ -0,0 +1,131 @@
...
+    run_command("sudo systemctl start adafruit_fan.service")
+    info('FAN', 'Done!')
+    print("You can stop the fan service with 'sudo systemctl stop adafruit_fan.service'")
+    print("You can start the fan service with 'sudo systemctl start adafruit_fan.service'")
+
+
+# Main function
+if __name__ == "__main__":
+    if os.geteuid() != 0:
+        print("Installer must be run as root.")
```
ladyada, line 128: nice

makermelissa:
It looks like clint is an external dependency, but it made the code much easier and shorter.

ladyada:
can you add a check for import failure and print out instructions on how to install (pip3 ...)

makermelissa:
Good idea. Added.

PR #126, opened 2020-10-19
increase txbuflen to 32k to accomodate bcm2835

jessa0:
According to the fbtft wiki: "To maximize throughput, make sure txbuflen is 16k or 32k (default is usually 4k). This is needed because of the DMA mapping/engine overhead per transfer that spi-bcm2835 has."

We could also make this configurable, but I don't know of any other reason for a user to change this value.

Only `minipitft13-overlay.dts` was tested, and on a Pi Zero W.


ladyada:
hihi what was the effect? like did you see it speed up?

geogriff:
> hihi what was the effect? like did you see it speed up?

Yes, it was about 15fps with stock 4k buffer, 30fps with increased 32k buffer. I didn't try to optimize between those two settings, since IIUC it's a single 32k buffer in RAM which is just a drop in a 512MB bucket.

Note that there are a few other modifications I had to make in conjuction with the `txbuflen` increase:
* Decrease the sleep time in `fbcp` from 33ms to 25ms
* Append `fps=60,speed=87500000` to the dtoverlay params for `minipitft13-overlay.dts` (I did not test how low I could set `speed`)
* Make sure VC4 3d acceleration is enabled and working.

makermelissa (approved):
Tested on the 1.14" display and it seems fine.

PR #137, opened 2021-01-08
Use hdmi rotation for 270 degrees

makermelissa:
This makes it so the braincraft hat can use the display upside down.


jepler:
This makes total sense, but I didn't test it yet.

jepler (approved):
Tested at 90 and 270, thanks!

PR #154, opened 2021-02-12
Dependency check for RPi.GPIO

eternaltinkering:
Dependency check for RPi.GPIO which is a dependancy for: Adafruit_Blinka/src/adafruit_blinka/microcontroller/bcm283x/pin.py

RPi.GPIO is not included in all Raspberry Pi OS images and may not be installed on the host E.G. Raspberry Pi OS Lite.


makermelissa (approved):
Thanks!

PR #167, opened 2021-03-14
Improve framerate, add joystick + buttons for 1.3" TFT bonnet

badcf00d:
The current usage for the 1.3" TFT Bonnet just uses the same config as the previous 1.3" MiniPiTFT, which works fine for the screen but it means the new joystick and buttons aren't mapped to anything.

I added a new install option for the TFT Bonnet with a new device tree that adds the buttons as gpio-keys. I also specified an fps for the devices that were missing them, because they seemed to be defaulting to a very slow refresh rate, something around 10fps. 

Regards,
Peter


makermelissa (approved):
Tested good. Thank you!

Issue #89, opened 2020-04-27
pitft-fbcp.sh failed on Ubuntu


windoze:
On RPi4B/Ubuntu-arm64, the script failed with the error
```
/home/ubuntu/repos/rpi-fbcp/main.c:9:10: fatal error: bcm_host.h: No such file or directory
```
This file is part of Raspbian, but seems it doesn't exist in Ubuntu.

Is there anyway I can use the screen in Ubuntu-arm64?'

Thanks.

ladyada:
not really - the software is tuned for raspbian kernel, if you find a solution you can submit a PR

craiglink:
```sudo apt-get install libraspberrypi-dev```

will fix the issue

ezraholm50:
Also had to install build-essential on DietPi. PR submitted.
Thanks.

makermelissa:
Closing as a bigger issue has eclipsed this and doesn't currently work with bookworm. See #295.

PR #168, opened 2021-03-15
Fix for FBCP fail on various debian based OS

ezraholm50:
libraspberrypi-dev & build-essential are needed for the compiling of FBCP. These are missing on some debian based operating systems. I'm running dietpi on a raspberrypi 3 and it failed to build.

Ubuntu mentioned, not heard back from issue owner https://github.com/adafruit/Raspberry-Pi-Installer-Scripts/issues/89


makermelissa (approved):
Tested that running this command on a pi didn't cause errors.

Issue #174, opened 2021-04-08
Maybe skip system upgrade in raspi-blinka.py 


timonsku:
I was wondering why the script does a full system upgrade?
I'm not sure if thats so clear for the users that this happens, this is a process that can potentially take hours and will touch huge parts of the system.
This can have really detrimental affects on a setup, I think this should at least be question if the user wants to do this or not instead of just doing it and forcing yes on every upgrade without the user seeing what happens.

If the intention was to upgrade to a newer Raspbian maybe instead have a check for the kernel version and only suggest a upgrade or re-install if its a very old version that can't be supported any more?

makermelissa:
Fixed by #377.

PR #178, opened 2021-05-14
Only do package update, no system upgrade

timonsku:
The guides already say to to an upgrade before even running this script and the way this works right now it just hangs for a long time without any feedback for sometimes 10-20 minutes.
This is I think not a very good UX. See this issue on more reasoning: https://github.com/adafruit/Raspberry-Pi-Installer-Scripts/issues/174
The default nature of upgrading all the system packages is I think a bit too many changes than one would expect from installing a python package.

### DIFF
--- a/raspi-blinka.py
+++ b/raspi-blinka.py
@@ -34,7 +34,7 @@ def check_blinka_python_version():
     if get_python3_version() < blinka_minimum_python_version:
         shell.bail("Blinka requires a minimum of Python version {} to install. Please update your OS!".format(blinka_minimum_python_version))
     
-def sys_update():   
+def sys_update():
     print("Updating System Packages")
     if not shell.run_command("sudo apt-get update"):
         shell.bail("Apt failed to update indexes!")
--- a/raspi-blinka.py
+++ b/raspi-blinka.py
@@ -38,9 +38,6 @@ def sys_update():
     print("Updating System Packages")
     if not shell.run_command("sudo apt-get update"):
         shell.bail("Apt failed to update indexes!")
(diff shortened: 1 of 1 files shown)


makermelissa:
The reason we have the upgrade is so we can have users having issues run the script to make sure they are up to date. I think we could add a parameter to skip the upgrade step. Would that be an acceptable compromise?

timonsku:
I think without feedback of the process I would not have it in there. On the weekend I ran it on a fresh Pi and apparently there was some error in the process and it was just stuck forever but you never really know.

If you think its really necessary to upgrade the whole system and not just the relevant packages I would have it ask during the process if you want to upgrade or not and if possible forward the output of apt to the user so they see what is going on and have them confirm individual things. At the moment it would just wipe all config changes a user may have made because of the -y option.

Otherwise I would maybe put in a notice but leave out the system upgrade for the user to run manually.

makermelissa:
Ok, yeah. Asking in the process is another great option.

makermelissa:
I updated the code to ask to update the packages rather than foregoing them altogether.

timonsku:
definitely better but can `adafruit_shell` maybe forward the stdin and stdout so that the -y option can be ommited?
Again I think overwriting custom configurations of system packages would be bad especially for beginner who might not know that an upgrade can do that.

makermelissa:
Kind of going through some of these that have been sitting awhile. regarding your last comment, adding stdin to python_shell is a great idea. I think it currently only does stdout and stderr. However, this might take some experimenting so it stays in order.

PR #180, opened 2021-05-27
Make i2smic.py noninteractive

beasteers:
We use i2smic in one of our ansible scripts and we need to be able to use it non-interactively, without requiring user intervention.

So we added cli flags that will force reboot/noreboot and autoload/noautoload

How we've used it:
```python
sudo python i2smic.py  -autoload -noreboot
```

This doesn't change default functionality btw.

NOTE: I replaced the use of `shell.prompt_reboot` because it doesn't allow for the use of `force_arg`


makermelissa (changes requested):
Maybe something simpler here would work better like:
```
    if not shell.argument_exists('noreboot'):
        shell.prompt_reboot()
    print("Exiting without reboot.")
    shell.exit()
```

beasteers:
That'd be great, except `shell.prompt_reboot()` doesn't let you pass `force_arg` so it's not fully interactive (even if that doesn't affect my use case). 

see: https://github.com/adafruit/Adafruit_Python_Shell/blob/master/adafruit_shell.py#L493-L501

If you're open to changes to the Adafruit Shell repo, a couple things that could clean this up would be


```python
# class Shell:
def prompt(self, message, *, default=None, force_arg=None, force_arg_no=True):
    if force_arg is not None and self.argument_exists(force_arg):
        return True
    if force_arg_no is True:  # default to using a 'no' prefix
        force_arg_no = 'no' + force_arg if force_arg else None
    if force_arg_no is not None and self.argument_exists(force_arg_no):
        return False
    ...

prompt('restart now?')  # no cli flags
prompt('restart now?', force_arg='reboot')  # supports: -reboot -noreboot
prompt('restart now?', force_arg='reboot', force_arg_no=None)  # supports: -reboot
prompt('restart now?', force_arg='reboot', force_arg_no='something')  # supports: -reboot -something


def prompt_reboot(self, force_arg='reboot', force_arg_no=True):
    if not self.prompt("REBOOT NOW?", default="y", force_arg=force_arg, force_arg_no=force_arg_no):
        ...
```

And then this pr would just be:
```
autoload = shell.prompt("Auto load module at boot?", force_arg="autoload")
...
shell.prompt_reboot()
```

makermelissa:
Hi, for some reason I wasn't getting what you were saying before and never got back to you. 

Upon re-reading this, I think adding , `**kwargs`to the `prompt_reboot` function in the shell library is all that would be required to make it behave like you want. I'll create a PR and release and then either of us can update this PR and get it merged in.

makermelissa:
https://github.com/adafruit/Adafruit_Python_Shell/pull/15 should fix it.

makermelissa:
Ok, I think I was still confused by how you wanted this to work. I updated shell to allow the `prompt` function to return a specific value if the argument exists.

makermelissa:
Code is now simplified and supports `reboot`, `noreboot`, `autoload`, and `noautoload` parameters. The "no" parameters have higher priority for the sake of simplicity, since you could technically pass both `noreboot` and `reboot` and likewise with `noautoload` and `autoload`.

makermelissa (approved):
It still works like you wanted, but is simpler now. Thanks.

PR #200, opened 2021-11-11
Added Pi Zero 2 w model

Margu86:
Added the new Pi Zero 2 w model.
Tested it on the new Board and works like a charm.


makermelissa (approved):
Thank you!

PR #192, opened 2021-09-09
Change backlight on ST7789 displays to be on by default

makermelissa:
Hard to believe the fix was so simple, but when you run out of documented suggestions that don't seem to work and just start trying your own wacky ideas, sometimes it just works out that way.
Fixes #187. Fixes #189.

PR #216, opened 2022-03-10
Updated backlight to be on for latest kernel

makermelissa:
Raspberry Pi updated the kernel again and now the backlight is off by default. This PR basically undos #192, which seems to fix it.


ladyada (approved):

notro:
This explains what happened with the backlight: https://github.com/notro/fbtft/issues/571#issuecomment-962970975

makermelissa:
Thanks, I saw that when I was working with the MIPI driver.

PR #220, opened 2022-04-14
fix(adafruit-pitft.py): libts dependency

NickEngmann:
Working with the Raspberry Pi Zero 2 W and ran into the error state where whenever I was attempting to install adafruit-pitft.py installer (`sudo python3 adafruit-pitft.py`) it would not find the libts dependency. Upon investigating, it seems like on Buster the library exists but is not indexable. 

This issue is solved by installing the libts-dev dependency.


makermelissa (approved):
Thanks

PR #234, opened 2022-11-12
Update readme, move scripts, bug fixes

makermelissa:
This fixes a few scripts and adds a lot of info to the README for folks who are directed from external sites that had shell scripts they were expecting moved on them.

@ladyada


ladyada (approved):
lgtm didnt test!

PR #236, opened 2022-11-30
Add Bug report template and additional conversions

makermelissa:
Add Bug report template to reduce asking for additional details. Apparently I hadn't pushed some of my conversions yet, so that's included.


brentru (approved):
@makermelissa  LGTM - Optionally you may want an "expected behavior" text area for the user to describe what _should_ happen.

makermelissa:
Thanks for reviewing. Yeah, I thought about that. I ended up taking the CircuitPython template as a starting point and it didn't have that, but we can always add it later.

PR #240, opened 2023-02-11
Update raspi-blinka.py

BasdeWinter:
fixes a "< operator cannot be used to compare NonType with int" error on a RPI-Zero-W


makermelissa (approved):
Thanks. Looks good to me. That shouldn't break anything.

PR #245, opened 2023-03-23
docs: add an alternative package if original fails

wise-introvert:
The old instructions mentioned adafruit-python**3**-shell as a dependency but when I tried to install the package I found out that the package didn't exist ( or deprecated I think ) and I was able to use adafruit-python-shell instead. So, I thought I might add the package to the instructions in case anyone else runs into the same issue as me.

### DIFF
--- a/libgpiod.py
+++ b/libgpiod.py
@@ -1,9 +1,7 @@
 # Instructions!
 # cd ~
 # sudo apt-get install python3-pip
-# sudo pip3 install --upgrade adafruit-python3-shell click 
-#   => NOTE: if the adafruit library fails, try installing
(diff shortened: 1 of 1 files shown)


makermelissa:
Oops, I think that was a typo. It should just be `adafruit-python-shell`. Want to just change it to only say adafruit-python-shell?

wise-introvert:
> Oops, I think that was a typo. It should just be `adafruit-python-shell`. Want to just change it to only say adafruit-python-shell?

Sure!

makermelissa (approved):
Awesome. Thank you.

PR #249, opened 2023-05-12
Change rotation to rotate for ST7789 displays

### LINKED ISSUE #241: rotation vs rotate
### Script Command

```shell
sudo script python3 adafruit-pitft.py --display=st7789_240x135 --rotation=270 --install-type=console
```

### Operating System

32-bit Buster Lite

### Hardware

Pi 4B

### Behavior

rotate works with the 1.14” Mini PiTFT in the /boot/config.txt, but rotation does not. 

### Description

A patch was added in November to the adafruit-pitft.py installer to use the word rotation. 

https://github.com/adafruit/Raspberry-Pi-Installer-Scripts/pull/233/files/2f8e05b9fc61a2f4be25d3579de02c02617bec99

This was probably a change compatible with newer kernels that Bullseye uses, but is not compatible with Buster. 

### Additional information

Adafruit Forum thread: https://forums.adafruit.com/viewtopic.php?p=960864

_No response_

makermelissa:
Fixes #241.

After some review, it looks like guide updates were not the right approach, but just changing how it writes the parameter into config.txt for the ST7789 displays only.


makermelissa:
Tagging @ladyada on this.

ladyada (approved):

PR #255, opened 2023-08-11
Please restore i2samp.h support for Raspbian buster system.

Lovmong:
Please restore i2samp.h support for Raspbian buster system. I test the i2s driver on the raspbian buster system, everything works fine.


makermelissa (approved):
Seems fair since you tested it.

Lovmong:
Thanks. 
It should be noted that speaker-test will only have sound when it is run with root privileges. Like: curl -sS https://raw.githubusercontent.com/adafruit/Raspberry-Pi-Installer-Scripts/master/i2samp.sh | sudo bash

PR #257, opened 2023-08-16
fix `click` install

mifi:
currently:

```
sudo apt-get install python3-pip click
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
E: Unable to locate package click
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,8 @@ We are in the process of converting our Shell Scripts to Python. Instructions fo
 To install the dependencies for the python scripts, run the following commands:
 
 ```bash
-sudo apt-get install python3-pip click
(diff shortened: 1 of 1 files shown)


makermelissa:
A lot of the scripts depend on click (https://pypi.org/project/click/) and haven't had a problem, however click isn't a package we have any control over. You may want to check your network connections and package sources as there isn't really anything to do on our end.

mifi:
Not sure I understand. This purpose of this PR is to install click using `pip3` instead of `apt`. Because I don't think `click` exists as a package in `apt`, hence I get the above error

ladyada (approved):

PR #262, opened 2023-10-13
Fix bug in python version check in raspi-blinka.py

Paulpanther:
The file checked the python version by comparing the major and minor versions. However, the variables compared were strings, making '11' > '8' -> False, when in fact it should have been true. I converted the versions to integers before comparing, so that versions 10 and above are accepted.


makermelissa (approved):
Thanks. that makes sense.

PR #266, opened 2023-10-20
Update raspi-config serial enable

caternuson:
Looks like the parameter for configuring serial non-interactively has changed. This PR updates it.


makermelissa (approved):
Looks good. Thanks.

PR #280, opened 2023-12-21
add ili9341 + tsc2007 overlay

timonsku:
Needs more testing when final hw is available.
Awaits Kernel PR for tsc2007 to be pulled into Pi Kernel repo here: https://github.com/raspberrypi/linux/pull/5791


makermelissa:
Ok, got the display working. Issue I was having is my device tree blob had some characters in it that the pi didn't like such as the underscore. Shortening to ili9341-backlight.dtbo fixed it. Touch doesn't work, but I presume it's because the PR you linked to hasn't been merged.

makermelissa:
Closing because it looks like this has already been superseded by #348.

PR #285, opened 2024-01-05
Add voice bonnet/braincraft hat script to ensure DTOs are enabled

makermelissa:
The voice bonnet driver install script doesn't correctly install on the Pi 5. I added a voice bonnet script that checks for certain issues and fixes them.

cc: @ladyada for review

### DIFF
--- a/voice_bonnet.sh
+++ b/voice_bonnet.sh
@@ -157,6 +157,19 @@ reconfig() {
         fi
 }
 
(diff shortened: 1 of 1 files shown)


makermelissa:
I submitted a PR (https://github.com/HinTak/seeed-voicecard/pull/23) to fix the actual voicecard install script, but this will work for now and continue to function once the voicecard install script is fixed.

ladyada (approved):

Issue #23, opened 2018-05-15
Script is stuck after "Configuring PiTFT..."


wromansky:
I've ran the script on the pi and it runs through the whole thing but stops at "Configuring PiTFT..." not sure what is going on.

ladyada:
hiya please post tech support Q's in the adafruit forums :)

PR #303, opened 2024-04-15
Fix #274

### LINKED ISSUE #274: I2S Mic may be out of date on Raspberry Pi OS Bookworm
Been trying to setup a pair of MEMS mics on my Raspberry Pi 400 using tutorials that make use of your setup script. Although it worked in the past, trying now it will run fine but does not show up as a recording device. I've asked around and they say its probably just out of date for current raspbian. If this could be updated to current or an alternative method given, that would be greatly appreciated.

LostNulide:
With Bookworm /boot/config.txt moved to /boot/firmware/config.txt.

This Commit depends on the Merge-Request #23 for the Adafruit_Python_Shell https://github.com/adafruit/Adafruit_Python_Shell/pull/23, for checking the current raspbian version.

### DIFF
--- a/i2smic.py
+++ b/i2smic.py
@@ -56,7 +56,10 @@ Installing...""")
         )
 
     # Enable I2S overlay
-    shell.run_command("sed -i -e 's/#dtparam=i2s/dtparam=i2s/g' /boot/config.txt")
+    if shell.get_raspbian_version() == "bookworm":
+        shell.run_command("sed -i -e 's/#dtparam=i2s/dtparam=i2s/g' /boot/firmware/config.txt")
+    else:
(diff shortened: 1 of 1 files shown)


makermelissa:
Hi, thanks for finding the issue and submitting a PR. In order to make this more futureproof, could you just check if the /boot/firmware/config.txt file exists? That way it will just work on whichever version is released in the next year or two.

Thanks.

LostNulide:
Hi,

i changed the code, to check for the config.txt, as you suggested.

You are welcome and thanks for the project :)

makermelissa (approved):
Thank you.

PR #311, opened 2024-05-03
drm added for PiTFT 3.5" resistive

mikeysklar:
The PiTFT 3.5" resistive display requires a drm flag to work with Pi Zero and 3B versions of Bookworm. Tested on Pi4 and Pi5 as well.

### DIFF
--- a/adafruit-pitft.py
+++ b/adafruit-pitft.py
@@ -118,7 +118,7 @@ config = [
                 "270": "touch-swapxy,touch-invy",
             },
         },
-        "overlay": "dtoverlay=pitft35-resistive,rotate={pitftrot},speed=20000000,fps=20",
+        "overlay": "dtoverlay=pitft35-resistive,rotate={pitftrot},speed=20000000,fps=20,drm",
 	    "overlay_drm_option": "drm",
         "force_x11": True,
         "calibrations": {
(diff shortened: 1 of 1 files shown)


makermelissa:
Unfortunately, this blanket approach will most likely break several other boards. Currently this is added in https://github.com/adafruit/Raspberry-Pi-Installer-Scripts/blob/main/adafruit-pitft.py#L426-L427 when `tinydrm_install` is set to **True** in `update_configtxt()`. This is called from https://github.com/adafruit/Raspberry-Pi-Installer-Scripts/blob/main/adafruit-pitft.py#L905 and is set to True if wayland is detected. It appears that they changed it in kernel 6.6 so that it can be bookworm, not wayland, and needs drm. So maybe trychanging line 905 to `if not update_configtxt(tinydrm_install=(not is_bullseye)):`

mikeysklar:
Okay, thank you for pointing this out. I'll make some changes and try the modified code on Bullseye and Bookworm.

mikeysklar:
Made suggested changes. Worked well on the following Pi's with a 3.5" resistive touch display.

Pi5 + 3/15/24 Bookworm + Updates
Pi3B + 3/15/24 Bookworm + Updates
Pi4 + Legacy Bullseye + Updates

Note Legacy Bullseye does not add the drm line to the dtoverlay. X and touch both work well without it.

makermelissa (approved):
Thanks, looks good.

PR #317, opened 2024-06-07
upstream: no longer use asoc_xxx() compatible macro

arrio464:
After linux 6.10-rc2, kernel remove asoc_xxx() compatible macro. So use original prototypes instead.


makermelissa (approved):
Thank you.

PR #327, opened 2024-09-23
Fix enable and disable of Wayland.

riktam:
raspi-config was being called with W1 as enable / disable parameter instead of 0 / 1 .


makermelissa (approved):
Thank you

Issue #331, opened 2025-01-29
adafruit-pitft.py : Programmer error, unrecognized boot option


mikeysklar:
### Script Command

```shell
sudo -E env PATH=$PATH python3 adafruit-pitft.py --display=35r --rotation=90 --install-type=console
```

### Operating System

Pi OS Bookworm 11/19/24 64-bit Desktop

### Hardware

Raspberry Pi 4B

### Behavior

adafruit-pitft.py installer script has started bailing since a set of updated packages were released this week. The error message is:

> Programmer error, unrecognized boot option

### Description

The lines that are causing this are from the raspi-config boot options:

```
#        if not shell.run_command("sudo raspi-config nonint do_wayland 0"):
#        if not shell.run_command("sudo raspi-config nonint do_wayland 1"):
```

Commenting them out did allow my Pi4B to run through the full adafruit-pitft.py and boot into console mode. 

Open forum issue with two other uses are different PiTFTs using Pi 3B are seeing this as well. All users of the adfafruit-pitft.py will be effected with the current Bookworm updates applied.

mikeysklar:
@makermelissa

mikeysklar:
It might just be a number change to raspi-config.

eg. 0,1 becomes W1, W2, W3 based on [raspi-config asciidoc](https://github.com/raspberrypi/documentation/blob/develop/documentation/asciidoc/computers/configuration/raspi-config.adoc#wayland-1)

console booting:
```
$ sudo raspi-config nonint do_boot_behaviour <B1/B2/B3/B4>

B1: boot to console, requiring login

B2: boot to console, logging in automatically

B3: boot to desktop, requiring login

B4: boot to desktop, logging in automatically
```

X11 vs Wayland booting:
```
$ sudo raspi-config nonint do_wayland <W1/W2/W3>

W1: use the Openbox window manager with X11 backend

W2: use the wayfire window manager with Wayland backend

W3: use the labwc window manager with Wayland backend
```

mikeysklar:
Yeah...minor argument renumbering. 

I'll drop in a PR. There might be more to add to it as I think this update could have broken other components.

makermelissa:
This was likely caused by https://github.com/adafruit/Raspberry-Pi-Installer-Scripts/pull/327.

mikeysklar:
Thx. Myself and another user verified #332 works, but there is some noise about various config files. I'll clean that up in a separate PR.

```
[...]

PR #332, opened 2025-01-29
bookworm update effects adafruit-pitft.py

mikeysklar:
Minor argument change to rapsi-config do_wayland setting. 

This is in regards to [issue# 331](https://github.com/adafruit/Raspberry-Pi-Installer-Scripts/issues/331) which is "Programmer error, unrecognized boot option" error message from adafruit-pitft.py since a January 2025 package update. raspi-config expect different arguments for do_wayland of W1, W2 and W3. I confirmed that all three work with a Pi 4B and 3.5" TFT.

### DIFF
--- a/adafruit-pitft.py
+++ b/adafruit-pitft.py
@@ -741,11 +741,11 @@ def disable_wayland(disable):
         return
     if disable:
         print("Using X11 instead of Wayland")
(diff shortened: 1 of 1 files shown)


makermelissa (approved):
Thanks, looks like the issues was caused by https://github.com/adafruit/Raspberry-Pi-Installer-Scripts/pull/327.

PR #336, opened 2025-02-25
Update RGB Matrix Installer

sciguy14:
This PR makes the following improvements/updates to the RGB Matrix Installer:
* Fixes the config.txt modifications to occur at the right place on bookworm builds (see https://github.com/RPi-Distro/raspberrypi-sys-mods/issues/88)
* Fixes python version checking that didn't actually appear to function correctly
* updates deprecated apt-get arguments to their modern equivalents
* Adds a setup option to automatically configure CPU core isolation on mutli-core Pis, per the recommendation from the RGB matrix repo maintainers
* Fully blacklists the kernel sound module instead of just changing the config.txt setting. This prevents complaints from the RGB matrix library on launch.
* Updates to the latest commit of the RGB matrix library, which I have tested and confirmed working on an Adafruit matrix+RTC hat.

### DIFF
--- a/rgb-matrix.sh
+++ b/rgb-matrix.sh
@@ -308,12 +308,17 @@ if [ $INSTALL_RTC -ne 0 ]; then
 fi
 
 if [ $QUALITY_MOD -eq 0 ]; then
-	# Disable sound ('easy way' -- kernel module not blacklisted)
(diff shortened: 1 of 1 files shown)


makermelissa:
superseded by #387

PR #346, opened 2025-06-17
Update rgb-matrix.sh

caternuson:
This PR updates the commit point used for the underlying [support library](https://github.com/hzeller/rpi-rgb-led-matrix) for RGB matrices. This library has no releases, so simply picking the latest commit point at the time of this PR.

It also makes the following significant changes:
* Removes Python 2 support. Must use Python 3 now.
* Adds additional `cython3` package needed for building Python bindings. 
* Only **builds** Python bindings but does **not install** them, since the make files do not support Python venv.
* Sound (`snd_bcm2835`) disabled via blacklist instead of using `config.txt`.
* Update path to `/boot/firmware` location used by latest RPi OS's.

### DIFF
--- a/rgb-matrix.sh
+++ b/rgb-matrix.sh
@@ -7,10 +7,11 @@
 # we reference a specific commit (update this as needed):
 GITUSER=https://github.com/hzeller
 REPO=rpi-rgb-led-matrix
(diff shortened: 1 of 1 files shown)


ladyada (approved):

ladyada:
yay!

Issue #368, opened 2026-01-19
Service aplay.service fails upon boot


besi:
### Script Command

```shell
/usr/bin/aplay -D default -t raw -r 44100 -c 2 -f S16_LE /dev/zero
```

### Operating System

Linux stereopida 6.12.47+rpt-rpi-v6 #1 Raspbian 1:6.12.47-1+rpt1~bookworm (2025-09-16) armv6l GNU/Linux

### Hardware

Raspberrypi Zero

### Behavior

I did run the speaker bonnet installation script as stated on the product page and I chose to install the anti-pop startup script.

However, upon boot I check the service and it tells me that it failed to start:

```
sudo systemctl  status aplay
× aplay.service - Invoke aplay from /dev/zero at system start.
     Loaded: loaded (/etc/systemd/system/aplay.service; enabled; preset: enabled)
     Active: failed (Result: exit-code) since Mon 2026-01-19 12:58:21 CET; 2min 34s ago
   Duration: 1.119s
    Process: 240 ExecStart=/usr/bin/aplay -D default -t raw -r 44100 -c 2 -f S16_LE /dev/zero (code=exited, status=1/FAILURE)
   Main PID: 240 (code=exited, status=1/FAILURE)
        CPU: 37ms
````

### Additional information

journalctl shows the following

```
Jan 19 12:58:20 stereopida aplay[240]: ALSA lib pcm_dmix.c:999:(snd_pcm_dmix_open) unable to open slave
Jan 19 12:58:21 stereopida aplay[240]: aplay: main:831: audio open error: No such file or directory
Jan 19 12:58:21 stereopida systemd[1]: aplay.service: Main process exited, code=exited, status=1/FAILURE
Jan 19 12:58:21 stereopida systemd[1]: aplay.service: Failed with result 'exit-code'.
```

makermelissa-ai-assistant:
Fixed by #369 (merged). The fix adds `After=sound.target` / `Wants=sound.target` ordering directives so the unit waits for ALSA to come up before starting, plus a bounded `Restart=on-failure` policy (`RestartSec=5s`, `StartLimitBurst=5`, `StartLimitIntervalSec=60`) for transient hiccups. Thanks @besi for the diagnosis and the PR!

makermelissa-ai-assistant:
Fixed by #369 (merged in 6358e0b). The fix combines:

[...]

PR #369, opened 2026-01-19
Configure aplay service to restart on failure

besi:
Add restart policy for aplay service

See #368

### DIFF
--- a/i2samp.py
+++ b/i2samp.py
@@ -105,6 +105,9 @@ pcm.!default {
 Description=Invoke aplay from /dev/zero at system start.
 
 [Service]
+Restart=on-failure
+RestartSec=5s
+
 ExecStart=/usr/bin/aplay -D default -t raw -r 44100 -c 2 -f S16_LE /dev/zero
 
 [Install]
--- a/i2samp.py
+++ b/i2samp.py
@@ -103,6 +103,10 @@ pcm.!default {
     shell.write_text_file("/etc/systemd/system/aplay.service", """
 [Unit]
 Description=Invoke aplay from /dev/zero at system start.
+After=sound.target
+Wants=sound.target
+StartLimitBurst=5
+StartLimitIntervalSec=60
 
 [Service]
 Restart=on-failure


makermelissa:
My concern with this addition is that it may cause a restart loop on the service if it keeps failing. Did you try restarting again after the failure and this was the only way to get it to start?

besi:
Once the Pi was started a manual restart did work without a problem. 
That's why I came up with the restart work around since it did the trick on my system.

However, maybe the service starts too soon upon boot and some pre-condition may be required instead? 
This could be a more clean solution.


```
sudo journalctl | grep error
Feb 04 14:08:12 stereopida aplay[242]: aplay: main:831: audio open error: No such file or directory
Feb 04 14:08:38 stereopida alsactl[426]: alsa-lib main.c:1541:(snd_use_case_mgr_open) error: failed to import hw:0 use case configuration -2
```

makermelissa-ai-assistant:
Thanks for digging in on this @besi — the journal output you posted is the key clue. `audio open error: No such file or directory` + the ALSA `hw:0` import failure both point at the service starting *before* ALSA has finished bringing up the card, not at `aplay` itself being flaky. So a bare `Restart=on-failure` will work, but it papers over the race rather than fixing it, and (as @makermelissa flagged) without retry limits it'll loop forever if the card is genuinely absent — log spam and a unit that's never idle on, e.g., a board with the HAT unplugged.

Could we combine both — order against `sound.target` so we wait for ALSA, and cap the retries so a real failure stops eventually? Something like:

```ini
[Unit]
Description=Invoke aplay from /dev/zero at system start.
After=sound.target
Wants=sound.target
StartLimitBurst=5
StartLimitIntervalSec=60

[Service]
Restart=on-failure
RestartSec=5s
ExecStart=/usr/bin/aplay -D default -t raw -r 44100 -c 2 -f S16_LE /dev/zero

[Install]
WantedBy=multi-user.target
```

That should:
- defer start until ALSA is plausibly ready (addresses the boot race),
- keep the restart cushion for transient hiccups,
- and give up after 5 failed starts in 60s instead of looping forever.

Happy to push a commit to your branch if you'd prefer, or you can update it.

makermelissa (changes requested):
I had an AI analysis done and agree with the findings.

makermelissa (approved):
Thank you

Issue #295, opened 2024-02-12
Adafruit Animated Eyes Bonnet for Raspberry Pi 5 (Bookworm)


mbandala:
The Animated Eyes Bonnet not working on Bookworm is [noted in the guide](https://learn.adafruit.com/animated-snake-eyes-bonnet-for-raspberry-pi/software-installation).

However, if you use a Python virtual environment on the Pi 5 (Legacy 32-bit OS Bookwork), the installation script should be able to work by modifying a few things. There are two issues. The FKMS overlay has some problems; eyes.py fails to properly set the resolution to 640x480, resulting in tiny eyes being displayed in 1080p. Even when attempting to manually set the text console resolution to 640x480, eyes.py still displays the eyes in 1080p. Conversely, the fbx2 program throws a "mmap() failed" error. I've attempted compiling fbx2, which compiles fine, but the error persists. Any assistance in resolving these two issues would be greatly appreciated. Thanks.

project-owner:
@makermelissa, will it also support Pygame 2 with this display - https://www.adafruit.com/product/2097?
https://forums.adafruit.com/viewtopic.php?t=208349

Thanks!

makermelissa:
The Pi_Eyes project won't, but I will be working on a single display version afterwards that should hopefully work on the Pygame. For now, you can use the [Easy Install](https://learn.adafruit.com/adafruit-pitft-3-dot-5-touch-screen-for-raspberry-pi/easy-install-2) method with the Mirrored Installation.

project-owner:
@makermelissa, where do you usually announce about your changes? On GitHub or Adafruit forum? Thanks!

Pyrogenez:
Did this ever get resolved? I'm noticing pi eyes says it doesn't work with bookworm but when running the scripts, the scripts use bookworm.

makermelissa:
Fixed by #373

PR #373, opened 2026-03-05
pi-eyes.sh trixie support

mikeysklar:
Pi OS has EOL'd Buster support. The [Snake Eyes Bonnet](https://learn.adafruit.com/animated-snake-eyes-bonnet-for-raspberry-pi/overview) had relied on a Pi 3B running Buster or a Pi 4B with experimental Bullseye support.

This PR gets resolves the FKMS (fake kernel mode setting) and uses KMS. This was the primary reason that pi-eyes got stuck before Bookworm. There are also quite a few updates to support modern Trixie installation changes. This resolves issue #295 

These are breaking changes not designed to support older releases. Only Trixie forward and brings adds the Pi 5 as as a hardware option now that it works with current Pi OS.

Outside of this PR the [fbx2.c](https://github.com/adafruit/Pi_Eyes/pull/24) driver will need to updated simultaniously in the [Pi Eyes repo](https://github.com/adafruit/Pi_Eyes).

This has been tested and works on:

  | Pi Model | OS                    | Display             |
  |----------|-----------------------|---------------------|
  | Pi 3B    | Trixie 32-bit Desktop | OLED 128x128 SSD1351|
  | Pi 4B    | Trixie 32-bit Desktop | OLED 128x128 SSD1351|
  | Pi 5     | Trixie 32-bit Desktop | OLED 128x128 SSD1351|

Summary of changes to pi-eyes.sh

  - Boot config path updated to /boot/firmware/
  - No longer swaps fkms overlay — leaves vc4-kms-v3d untouched
  - gpu_mem=128 written directly to config.txt
  - pip installs into Python venv (PEP 668)
  - Code installs to /opt/Pi_Eyes instead of /boot
  - Dropped do_overscan and libatlas-base-dev (removed from Trixie)
  - rpi-lgpio used on Pi 5 instead of RPi.GPIO
  - systemd units replace rc.local
  - Always uses xinit to launch eyes.py
  - Forces HDMI on headless via video=HDMI-A-1:640x480@60e

### DIFF
--- a/pi-eyes.sh
+++ b/pi-eyes.sh
@@ -1,370 +1,384 @@
 #!/bin/bash
+# Adafruit Snake Eyes Bonnet installer for Raspberry Pi OS Trixie (Debian 13).
+# Supports Pi 3B, Pi 4, and Pi 5.
+#
+# Changes vs upstream pi-eyes.sh:
+#   - Targets Trixie only; drops Buster/Bullseye legacy paths
+#   - Boot config always at /boot/firmware (Trixie standard)
+#   - Does not touch vc4-kms-v3d overlay (fkms was removed in Bookworm)
+#   - Forces HDMI mode via video= kernel cmdline (works headless)
+#   - pip installs into Python venv (PEP 668 / externally-managed-env)
+#   - Code installed to /opt/Pi_Eyes (not /boot — FAT32 partition)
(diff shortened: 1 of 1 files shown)


mikeysklar:
@makermelissa @ladyada

ladyada (approved):

PR #375, opened 2026-04-23
i2samp: fix Speaker Bonnet audio on Bookworm/Trixie

mikeysklar:
Two fixes to `i2samp.py` for Speaker Bonnet on Bookworm/Trixie:

- Bind `asound.conf` to the ALSA card name (`sndrpigooglevoi`) instead of index 0, so `speaker-test -c2` routes to the Bonnet regardless of what other audio devices enumerate first.
- Comment out `dtparam=audio=on` in `config.txt` — the installer never stripped it, leaving built-in audio active and conflicting with the I2S overlay.

Tested on Pi 4 / Trixie 64-bit Lite with Speaker Bonnet

### DIFF
--- a/i2samp.py
+++ b/i2samp.py
@@ -13,8 +13,9 @@ PRODUCT_NAME = "I2S Amplifier"
 OVERLAY = "googlevoicehat-soundcard"
 # ALSA card id registered by the overlay above (see /proc/asound/cards).
 # Binding asound.conf by name rather than numeric index keeps routing
-# correct when HDMI/USB audio pushes the I2S card off card 0.
-CARD_NAME = "sndrpigooglevoicehat"
+# correct when HDMI/headphones/USB audio pushes the I2S card off card 0.
+# Note: the kernel truncates ALSA card short names to 15 chars.
+CARD_NAME = "sndrpigooglevoi"
 
 def driver_loaded(driver_name):
     return shell.run_command(f"lsmod | grep -q '{driver_name}'", suppress_message=True)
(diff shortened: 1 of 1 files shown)


makermelissa (commented):
@mikeysklar I noticed "sndrpigooglevoi" is hard-coded. You mentioned this is "sndrpigooglevoicehat" truncated to 15 characters, but I'm not sure where you got that name from. In order to help future-proof this, is it possible to update the script to get the name from whereever you got it from and use CARD_NAME[:15] to truncate it? Also, could you check if this affect the MAX98357a breakout?

mikeysklar:
The I2S device isn't enumerated until after the overlay is loaded and the Pi reboots, so `aplay -l` can't be queried on initial install. Making it tricky to dynamically pick up the name. Maybe you have a suggestion for how to get the name pre-reboot? Maybe `dtoverlay -l`?

ALSA has its own rename process I've been going with.

kernel driver uses `snd_rpi_googlevoicehat_soundcard`
ALSA strips out underscores and drops it to 15 chars -> sndrpigooglevoi

Updated the hard coded `CARD_NAME` with the full name and `[:15]` slice so the truncation is at least explicit. 

Will re-test on Speaker Bonnet with the change. 

I don't have the MAX98357A breakout (out of stock) but same driver/chip as the Speaker Bonnet. I can request the out-of-stock breakout board through CSP if you think it is important to verify.

makermelissa:
Ok thanks. If it tests good, then we can leave it here for now.

mikeysklar:
During testing I realized that even the reboot isn't critical. We can just load the overlay at the end of the install script.

`sudo dtoverlay googlevoicehat-soundcard`

Then `aplay -l` can discover the card id dynamically and pull the card name.

Falls back to `sndrpigooglevoi` on failed discovery.

Works with Pi OS Trixie Desktop and Lite 64-bit builds. Tested with Speaker Bonnet on a Pi 4.

makermelissa (approved):
Perfect. Thank you.

Issue #341, opened 2025-03-30
32-bit ST7789 1.3" 240x240 Partial Screen on boot


mikeysklar:
### Script Command

```shell
sudo -E env PATH=$PATH python3 adafruit-pitft.py --display=st7789v_bonnet_240x240 --rotation=0 --install-type=console
```

### Operating System

Pi OS Bookworm, Lite, 32-bit

### Hardware

Pi Zero W

### Behavior

The [Adafruit 1.3" 240x240 Color TFT+Joystick Bonnet](https://www.adafruit.com/product/4506) boots up without showing the init scripts and the initial login shell is missing 20% of the right side of the screen. 

### Description

The adafruit-pitft.py installer works well in getting a usable screen and subsequent commands run after boot are full screen. Is there a way to get the startup init script output and correct screen size draw on boot?

![Image](https://github.com/user-attachments/assets/31e3e325-b9df-42d3-9c60-819d8ffb5a12)
![Image](https://github.com/user-attachments/assets/26c261a3-c877-455a-baeb-f785b55f9210)

### Additional information

Using the same bonnet with a 64-bit install on a Pi Zero 2W works well. Init scripts are shown and screen boot is correct resolution. The problem is tied to 32-bit Pi OS releases both Bookworm and Bullseye show the same behavior. 

[Forum issue](https://forums.adafruit.com/viewtopic.php?t=217292) for reference.

grandinquisitor:
I noticed this

```
sudo dmesg | grep -i fb
[    2.641552] simple-framebuffer 1eaf0000.framebuffer: fb0: simplefb registered!
[   41.538857] panel-mipi-dbi-spi spi0.0: [drm] fb0: panel-mipi-dbid frame buffer device
```

which makes it look like it's loading the driver really late in the boot process

makermelissa:
Reopening. I noticed it is still doing it. What I did notice is it looks good just until it gets to the command prompt and then suddenly the right side turns black.

PR #378, opened 2026-05-20
Fix PiTFT race condition between vc4-kms-v3d and SPI TFT (#365)

### LINKED ISSUE #365: PiTFT Plus does not work on 50% of boot cycles, probably due to race condition (/dev/fb0 vs. /dev/fb1)
### Script Command

```shell
git clone https://github.com/adafruit/Raspberry-Pi-Installer-Scripts.git

cd Raspberry-Pi-Installer-Scripts

sudo /home/pi/.pyenv/shims/python adafruit-pitft.py --display=28c --rotation=90 --install-type=console --reboot no
```

### Operating System

Raspberry Pi OS Trixie Lite (64-bit)

### Hardware

Raspberry Pi Zero 2 W and Adafruit PiTFT Plus (2.8" capacitive touch)

### Behavior

On 50% of boot cycles, the PiTFT Plus shows the normal output from initialization and works fine. During these boot cycles, /dev/fb0 exists and /dev/fb1 does not exist.

On the other 50% of boot cycles, the PiTFT Plus is inaccessible and remains in its state at power-on (blank screen, output from previous boot cycle, etc.) During these boot cycles, /dev/fb1 exists and /dev/fb0 does not exist.

This behavior is reproducible: repeatedly booting the Raspberry Pi Zero 2 W resulted in one of these two states. There does not appear to be a pattern or any clear reason for ending up in one state or the other.

### Description

Boot cycles that result in a functional PiTFT contain the following lines in journalctl and dmesg:

```
simple-framebuffer 1eaf0000.framebuffer: framebuffer at 0x1eaf0000, 0x10a800 bytes
simple-framebuffer 1eaf0000.framebuffer: format=a8r8g8b8, mode=656x416x32, linelength=2624
Console: switching to colour frame buffer device 82x26
simple-framebuffer 1eaf0000.framebuffer: fb0: simplefb registered!

...

[drm] Initialized ili9341 1.0.0 for spi0.0 on minor 1
Console: switching to colour frame buffer device 40x30
ili9341 spi0.0: [drm] fb0: ili9341drmfb frame buffer device
```

Boot cycles that result in a non-functional PiTFT contain the following lines in journalctl and dmesg:

```
simple-framebuffer 1eaf0000.framebuffer: framebuffer at 0x1eaf0000, 0x10a800 bytes
simple-framebuffer 1eaf0000.framebuffer: format=a8r8g8b8, mode=656x416x32, linelength=2624
Console: switching to colour frame buffer device 82x26
simple-framebuffer 1eaf0000.framebuffer: fb0: simplefb registered!

...

[drm] Initialized ili9341 1.0.0 for spi0.0 on minor 0
ili9341 spi0.0: [drm] fb1: ili9341drmfb frame buffer device
```

Attached are the journalctl and dmesg dumps for example nonworking and working boot cycles.

[working.txt](https://github.com/user-attachments/files/23394565/working.txt)

[nonworking.txt](https://github.com/user-attachments/files/23394568/nonworking.txt)

### Additional information

Upon further investigation, I traced the output in the latter chunk of journalctl/dmesg output above to /usr/local/bin/con2fbmap-helper.sh, which contains this:

```
# Wait up to 30 seconds for /dev/fb0 or /dev/fb1 to appear
for i in {1..300}; do
    for fbdev in 0 1; do
        if [ -e /dev/fb$fbdev ]; then
            echo "Found /dev/fb$fbdev, checking if it's ili9341..."

            # Check if it's actually the ili9341 device
            if dmesg | grep -q "ili9341.*fb$fbdev"; then
                echo "ili9341 framebuffer ready, mapping console..."
                con2fbmap 1 $fbdev
                echo "Console mapped to framebuffer $fbdev"
                exit 0
            fi
        fi
    done
    sleep 0.1
done
```

The problem looks clear to me: in both cases, initialization begins with a mapping to fb0, but then con2fbmap-helper.sh waits for the display device (ili9341) to appear as either fb0 or fb1. If it happens to appear as fb0, then configuration proceeds normally and the display works. But if it happens to appear as fb1, then the kernel attempts to configure the display on fb1, which does not match up with the earlier device reservations at fb0.

The arbitrary nature of the mapping looks a lot like a race condition. I reviewed the preceding lines in journalctl and dmesg for working and non-working configurations (particularly those after the simplefb initialization and the con2fbmap-helper.sh output), but could not find any reason why the display device variously appears as fb0 or fb1. It's entirely possible that there is some difference, but I can't find any.

makermelissa-ai-assistant:
Fixes the PiTFT race condition reported in #365 where the SPI TFT shows nothing on ~50% of boot cycles.

## What was going wrong

`con2fbmap-helper.sh` discovered the SPI TFT's framebuffer number by greping `dmesg` for a line like `{display_type}.*fb<n>`. That worked when ili9341 won the probe race against `vc4-kms-v3d` and landed on `/dev/fb0`, but failed when vc4 won and ili9341 landed on `/dev/fb1`:

1. **The grep is brittle.** `dmesg` output is buffered/rate-limited, and `ili9341.*fb0` matches any line where those tokens appear in that order, not specifically the fbdev registration message.
2. **`con2fbmap 1 N` doesn't repaint.** It only updates the console→fb routing table. By the time the helper runs at `multi-user.target`, getty has already painted the login prompt to the old fb (vc4's HDMI surface). Nothing writes to tty1 again, so the SPI display stays at whatever was in its panel RAM at power-on — exactly the symptom the reporter described.

## Fix

- Walk `/sys/class/graphics/fb*` and use the authoritative `name` attribute plus the resolved `device` sysfs path to identify the SPI TFT, instead of greping `dmesg`. Numeric fb entries only (so `fbcon` is skipped).
- Match by either:
  - **fb name substring** — covers DRM tinydrm (`ili9341drmfb`, `hx8357drmfb`, etc.) and legacy fbtft staging (`fb_ili9341`, …).
  - **Device path contains `/spi0.0/`** — covers `panel-mipi-dbi-spi` (fb name is just `mipi_dbi`), with `vc4drmfb` and `simple-framebuffer` explicitly excluded as a belt-and-braces guard.
- After `con2fbmap 1 N`, send `ESC c` (terminal reset) to `/dev/tty1` to force fbcon to repaint the active console on the newly-mapped framebuffer. This is the missing piece that makes the display actually show content even when `con2fbmap` resolves the correct fb.
- On timeout, log every framebuffer the helper saw (name + device path) for easier debugging in the future.

## Testing

Smoke-tested the matcher against the cases that matter:

| display_type | fb name | device path | expected | result |
|---|---|---|---|---|
| `ili9341` | `ili9341drmfb` | `.../spi0.0/drm/card0` | match | ✅ |
| `ili9341` | `fb_ili9341` (fbtft) | `.../spi0.0/graphics/fb1` | match | ✅ |
| `hx8357d` | `hx8357drmfb` | `.../spi0.0/drm/card0` | match | ✅ |
| `panel-mipi-dbi-spi` | `mipi_dbi` | `.../spi0.0/drm/card0` | match | ✅ |
| `ili9341` | `vc4drmfb` | `.../3f902000.hdmi/...` | no match | ✅ |
| `ili9341` | `simple-framebuffer` | `.../1eaf0000.framebuffer` | no match | ✅ |

Also confirmed `bash -n` is clean and the timeout/diagnostic-dump path works when no SPI TFT is attached.

Closes #365.

### DIFF
--- a/templates/con2fbmap-helper.sh
+++ b/templates/con2fbmap-helper.sh
@@ -1,25 +1,107 @@
 #!/bin/bash
 
-# Wait for TFT framebuffer to be ready
-echo "Waiting for SPI TFT framebuffer..."
+# Map the Linux console (tty1) to the SPI TFT framebuffer.
+#
+# Background: when vc4-kms-v3d (HDMI) and the SPI TFT DRM driver
+# (ili9341 / hx8357 / mipi-dbi-spi) both load at boot, they race for
+# DRM minor / framebuffer numbers. The SPI display can land on either
+# /dev/fb0 or /dev/fb1 depending on probe order. Greping dmesg for a
+# fixed fb number is brittle (issue #365), so we discover the SPI TFT
+# framebuffer by walking /sys/class/graphics/fb*.
+#
+# After remapping, we force a console refresh so whatever was last
+# drawn on the old framebuffer gets re-rendered on the SPI display.
 
-# Wait up to 30 seconds for /dev/fb0 or /dev/fb1 to appear
-for i in {1..300}; do
-    for fbdev in 0 1; do
-        if [ -e /dev/fb$fbdev ]; then
-            echo "Found /dev/fb$fbdev, checking if it's {display_type}..."
+set -u
 
-            # Check if it's actually the ili9341 device
-            if dmesg | grep -q "{display_type}.*fb$fbdev"; then
-                echo "{display_type} framebuffer ready, mapping console..."
-                con2fbmap 1 $fbdev
-                echo "Console mapped to framebuffer $fbdev"
-                exit 0
-            fi
+DISPLAY_TYPE="{display_type}"
+TIMEOUT_DECISECONDS=300   # 30 seconds (loop sleeps 0.1s)
+
+echo "Waiting for SPI TFT framebuffer (display_type=${DISPLAY_TYPE})..."
+
+# Return 0 (and echo the fb number) if /sys/class/graphics/$1 looks
+# like the SPI TFT we installed for.
+is_spi_tft_fb() {
+    local fbpath="$1"
+    local fbname
+    local devpath
+
+    # Read the framebuffer driver name, e.g. "ili9341drmfb", "hx8357drmfb",
+    # "mipi_dbi", "vc4drmfb", "simple-framebuffer".
+    fbname="$(cat "${fbpath}/name" 2>/dev/null || true)"
+
+    # Resolve the underlying device path (DRM card → parent device).
+    devpath="$(readlink -f "${fbpath}/device" 2>/dev/null || true)"
+
+    # 1) Name-based match. Works for ili9341, hx8357d, st7789, etc. The
+    #    fbdev name varies by driver: DRM tinydrm uses "<name>drmfb"
+    #    (e.g. "ili9341drmfb"); the older fbtft staging driver uses
+    #    "fb_<name>" (e.g. "fb_ili9341"). Accept either.
+    if [ -n "${fbname}" ]; then
+        if [[ "${fbname}" == *"${DISPLAY_TYPE}"* ]]; then
+            return 0
+        fi
+    fi
+
+    # 2) Device-path match. For panel-mipi-dbi-spi the fbdev name is
+    #    "mipi_dbi" which does not contain the configured display_type,
+    #    but the device path under sysfs lives below the SPI controller
+    #    (e.g. .../soc/.../spi@.../spi0.0/...). Match any SPI slave on
+    #    spi0.0, which is where the installer wires SPI TFT panels.
+    if [ -n "${devpath}" ] && [[ "${devpath}" == *"/spi0.0/"* ]]; then
+        # Belt-and-braces: exclude framebuffers that are obviously not
+        # the SPI TFT (e.g. vc4drmfb, simple-framebuffer). Those should
+        # not have spi0.0 in their devpath, but guard anyway.
+        if [[ "${fbname}" != "vc4drmfb" && "${fbname}" != "simple-framebuffer" ]]; then
+            return 0
+        fi
+    fi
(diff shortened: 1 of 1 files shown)


makermelissa-ai-assistant:
Hey @neuron-whisperer — fix candidate is up. Two changes vs. current `main`:

1. Helper now discovers the SPI TFT framebuffer via `/sys/class/graphics/fb*/name` + sysfs device path (matches `/spi0.0/`) instead of greping `dmesg`. Should reliably find the panel regardless of which fb number it landed on.
2. After `con2fbmap 1 N`, sends `ESC c` to `/dev/tty1` to force fbcon to repaint. This is what I think is actually fixing the blank-screen symptom — `con2fbmap` only updates the routing table, it doesn't redraw, and getty had already painted onto the wrong fb before the helper ran.

If you have time to test on your Pi Zero 2 W + PiTFT Plus 2.8" cap-touch rig:

```bash
git clone -b fix/365-con2fbmap-race \
  https://github.com/makermelissa-piclaw/Raspberry-Pi-Installer-Scripts.git
cd Raspberry-Pi-Installer-Scripts
sudo python3 adafruit-pitft.py --display=28c --rotation=90 \
  --install-type=console --reboot no
sudo reboot
```

Then reboot ~15-20 times and confirm the display comes up every time. If you want to see which fb the panel landed on each boot, `cat /sys/class/graphics/fb*/name` after login — should see `ili9341drmfb` on either fb0 or fb1, and either way the console should be visible on the TFT.

If a boot ever shows blank panel + working SSH login, `systemctl status con2fbmap.service` output would be gold — it now dumps every fb it saw on timeout.

Thanks for the detailed repro in the issue, by the way — saved a lot of guesswork.

makermelissa:
I'm going to merge this because it would likely help with #341.

makermelissa (approved):

PR #384, opened 2026-05-25
Gate raspi-config calls behind Shell.run_raspi_config() (fixes #321)

### LINKED ISSUE #321: adafuit-pitft.py fails when using Dietpi linux distro
### Script Command

```shell
sudo -E env PATH=$PATH python3 adafruit-pitft.py --display=35r --rotation=90 --install-type=fbcp
```

### Operating System

32bit DietPi v9.6.1

### Hardware

Raspberry Pi Model B Rev 2

### Behavior

Errors about "is_wayland"   Currently installing, can edit this with precise errors afterward.

### Description

After doing some digging I discovered that the issue is with this line:  `output = shell.run_command("loginctl show-session $(loginctl | grep $(whoami) | awk '{print $1}') -p Type | grep wayland", suppress_message=True, return_output=True, run_as_user=username).strip()`

It appears that DietPi doesn't include `loginctl`.  I'm able to bypass the check by manually editing the script to hard code it to "x11".   

It would be great if I was able to bypass detection, maybe by passing an argument when running the script  e.g. `sudo -E env PATH=$PATH python3 adafruit-pitft.py --display=35r --display-manager=x11 --rotation=90 --install-type=fbcp`

Cheers!

### Additional information

_No response_

makermelissa-ai-assistant:
## Summary

The Python installer scripts in this repo shell out to `raspi-config nonint ...` to toggle SPI/I2C, set overscan, change boot behaviour, configure the fan service, etc. `raspi-config` is only shipped (and only honored) on Raspberry Pi OS — on DietPi, Ubuntu, and other Debian-based distros the binary is absent and `shell.run_command` returns falsy on the missing executable, so the tweak silently doesn't happen and the user sees a confusing `command not found` line mid-install. That's the underlying "non-Pi-OS distro" gap behind #321; the reporter's specific `is_wayland()`/`loginctl` complaint was already fixed when that helper was deleted in d14339d, but the broader raspi-config-unavailable problem on the same distro remained.

## Approach

[adafruit/Adafruit_Python_Shell#38](https://github.com/adafruit/Adafruit_Python_Shell/pull/38) (now merged + released) added a reusable `Shell.run_raspi_config(args, suppress_message=False, return_output=False, run_as_user=None)` that gates on `is_raspberry_pi_os()` and short-circuits to a no-op on non-Pi-OS hosts (returning `True`, or `""` when `return_output=True` so `.strip()` callers stay happy).

This PR migrates every `raspi-config nonint` call site in the repo's Python installers to use it.

| File | Sites | What they toggle |
| --- | --- | --- |
| `adafruit-pitft.py` | 6 | `do_boot_behaviour`, `do_boot_splash`, `do_boot_target`, `do_overscan` ×2 |
| `raspi-blinka.py` | 7 | `do_i2c`, `do_spi`, `do_serial_hw`, `do_serial`, `do_ssh`, `do_camera`, `disable_raspi_config_at_boot` |
| `raspi-spi-reassign.py` | 3 | `do_spi` ×2 + `get_spi` (uses `return_output=True` + `.strip()`) |
| `joy-bonnet.py` | 2 | `do_i2c`, `do_overscan` |
| `retrogame.py` | 1 | `do_i2c` |
| `pitft-fbcp.py` | 1 | `do_spi` |
| `rtc.py` | 1 | `do_i2c` |
| `adafruit_fanservice.py` | 1 | `do_fan` — was previously gated only on `is_raspberry_pi()`, which is true on DietPi-on-Pi-hardware |

Total: **22 call sites across 8 files**, +22 / −22 net.

The leading `sudo` on the old `run_command` lines is dropped: all of these installers already require root (`require_root()` or equivalent), so the inner sudo was redundant.

## Out of scope

Bash-only scripts in this repo (`pi-eyes.sh`, `arcade-bonnet.sh`, `rgb-matrix.sh`, `retrogame.sh`, `spectro.sh`) still call `raspi-config` directly. They don't use `adafruit_shell` and would need a different fix; happy to follow up separately if you want that too.

## Test plan

- Syntax (`ast.parse`) clean across all 8 files (pre-existing `SyntaxWarning`s on invalid escape sequences are unrelated).
- Confirmed via `ast` walk that all **22** migrated call sites resolve to syntactically valid `shell.run_raspi_config(...)` invocations, with `suppress_message` / `return_output` kwargs preserved on the three sites that use them.
- On this Pi (Raspberry Pi OS), with `Shell.run_command` monkey-patched to record calls: `run_raspi_config("do_overscan 1")` forwards to `run_command("raspi-config nonint do_overscan 1", ...)` exactly as before.
- On simulated non-Pi-OS (`is_raspberry_pi_os` patched to `False`): all calls short-circuit, return truthy / empty string, zero `run_command` invocations.

## References

- Closes #321.
- Depends on adafruit/Adafruit_Python_Shell#38 (merged + released).

### DIFF
--- a/adafruit-pitft.py
+++ b/adafruit-pitft.py
@@ -595,7 +595,7 @@ def install_console():
 
     print("Setting raspi-config to boot to console w/o login...")
     shell.chdir(target_homedir)
-    shell.run_command("raspi-config nonint do_boot_behaviour B2")
+    shell.run_raspi_config("do_boot_behaviour B2")
 
     # remove fbcp
     shell.pattern_replace("/etc/rc.local", "^.*fbcp.*$")
@@ -608,10 +608,10 @@ def uninstall_console():
     if is_desktop:
         print("Restoring Desktop Environment...")
         shell.run_command("apt-get -y install rpd-plym-splash")     # Install Splash Screen
-        shell.run_command("raspi-config nonint do_boot_splash 0")   # Enable Splash Screen
+        shell.run_raspi_config("do_boot_splash 0")                  # Enable Splash Screen
         if not shell.is_minimum_version("trixie"):
             shell.run_command("apt-get -y install raspberrypi-ui-mods") # Reinstall Raspberry Pi OS UI mods
-        shell.run_command("raspi-config nonint do_boot_target B2")  # Boot to Desktop
+        shell.run_raspi_config("do_boot_target B2")                 # Boot to Desktop
 
     if shell.exists("/etc/systemd/system/con2fbmap.service"):
         print("Removing console fbcon map service...")
@@ -678,10 +678,10 @@ def install_mirror():
     # if desktop environment is installed...
     if is_desktop:
         print("Setting raspi-config to boot to desktop w/o login...")
-        shell.run_command("raspi-config nonint do_boot_behaviour B4")
+        shell.run_raspi_config("do_boot_behaviour B4")
 
     # Disable overscan compensation (use full screen):
-    shell.run_command("raspi-config nonint do_overscan 1")
+    shell.run_raspi_config("do_overscan 1")
     # Set up HDMI parameters:
     print("Configuring boot/config.txt for forced HDMI")
     shell.reconfig(f"{boot_dir}/config.txt", "^.*hdmi_force_hotplug.*$", "hdmi_force_hotplug=1")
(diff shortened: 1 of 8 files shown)


makermelissa-ai-assistant:
Follow-up: I've also opened [adafruit/Adafruit_Python_Shell#38](https://github.com/adafruit/Adafruit_Python_Shell/pull/38) which adds the same gating logic as a reusable `Shell.run_raspi_config()` method, plus migrates the existing `Shell.set_window_manager()` call internally.

Once that lands and is released to PyPI, the local `run_raspi_config()` helper in this PR can be deleted in favor of `shell.run_raspi_config(...)`, and the same migration can be applied to the other call sites in this repo (`raspi-blinka.py`, `raspi-spi-reassign.py`, `joy-bonnet.py`, `retrogame.py`, `pitft-fbcp.py`) — they all currently shell out to `raspi-config nonint ...` the same way.

Happy to do this either way:
- **Land this PR as-is** (self-contained local helper, ships immediately), then do a follow-up PR after the Shell release to migrate all six scripts to the library helper.
- **Hold this PR** until Shell #38 ships, then rewrite to depend on the new method directly.

Let me know which you'd prefer.

makermelissa (approved):
Looks good to me.

PR #386, opened 2026-05-27
adafruit-pitft: fix overlay_params lookup (fixes #340)

### LINKED ISSUE #340: 2.8" PiTFT inverse touch points
### Script Command

```shell
sudo -E env PATH=$PATH python3 adafruit-pitft.py --display=28c --rotation=90 --install-type=mirror

sudo -E env PATH=$PATH python3 adafruit-pitft.py --display=28r --rotation=90 --install-type=mirror
```

### Operating System

PiOS Bookworm
Desktop
64-bit

### Hardware

Raspberry Pi 4B
* [2.8" PiTFT (resistive)](https://www.adafruit.com/product/2298)
* [2.8" PiTFT (capacitive)](https://www.adafruit.com/product/2423)

### Behavior

inverted x,y coordinates on touchscreen for both models of display.

### Description

adafruit-pitft.py brings up a visually correct X11 session, but the touch points are inverted. Bottom right corner touch results in top left corner touch. This was shown to happen on both capacitive and resistive 2.8" PiTFTs. The workaround is to change invert x,y at the config.txt level.

```
touch-invx,touch-invy
```

capacitive
```
dtoverlay=pitft28-capacitive,rotate=90,speed=64000000,fps=30,touch-invx,touch-invy,drm
```

```
dtoverlay=pitft28-resistive,rotate=90,speed=64000000,fps=30,touch-invx,touch-invy,drm
```

This had been working and I'm not sure when it stopped. 

[forum issue](https://forums.adafruit.com/viewtopic.php?t=217342)

### Additional information

_No response_

makermelissa-ai-assistant:
## Summary

Fixes #340 — touch coordinates inverted at `rotation=90` (and effectively unconfigured at every rotation) for PiTFT mirror installs.

## Root cause

`update_configtxt()` reads `overlay_params` from the top level of `pitft_config`, but the data is nested under `pitft_config["touchscreen"]["overlay_params"]`. The `if` is always `False`, so the `touch-invx` / `touch-invy` / `touch-swapxy` parameters never get appended to the `dtoverlay` line — for **any** PiTFT display, at **any** rotation, in mirror install mode.

This was introduced in commit 34817dc ("Fix capacitive touch rotations", 2024) — the same commit that added the per-rotation `overlay_params` data structure. The values were placed under `touchscreen`, but the lookup was written against the top level. Symptoms were subtle enough (touch only inverted at specific rotations, which calibration tools partially mask) that the bug went unnoticed.

## Fix

One-liner: read `overlay_params` from `pitft_config["touchscreen"]["overlay_params"]` where it actually lives. The patch also tidies the lookup into three clear steps and adds a comment pointing at this issue for future readers.

## Verification

Tested on real hardware: Raspberry Pi 4 + Adafruit 2.8" PiTFT capacitive (28c), Raspberry Pi OS Trixie Desktop 64-bit, kernel 6.12.

**Before the fix:**
```
$ grep pitft28 /boot/firmware/config.txt
dtoverlay=pitft28-capacitive,rotate=90,speed=64000000,fps=30,drm
```
No `touch-*` params — matches the reporter's symptoms exactly.

**After the fix:**
```
$ grep pitft28 /boot/firmware/config.txt
dtoverlay=pitft28-capacitive,rotate=90,speed=64000000,fps=30,drm,touch-swapxy,touch-invy
```

Then captured `evtest` events while tapping the four corners of the display in its rotated (landscape) orientation:

| Corner | ABS_X (0–319) | ABS_Y (0–239) |
|---|---|---|
| Top-left | 49 | 71 |
| Top-right | 293 | 78 |
| Bottom-left | 65 | 214 |
| Bottom-right | 258 | 207 |

Axes track the physical display correctly: TL→TR raises X, TL→BL raises Y. ✅

## Scope notes

- The fix applies to all PiTFT models that use `overlay_params` (28r, 28c, 35r). Only 28c was verifiable on my bench, but the lookup is uniform across displays so 28r/35r should be unblocked by the same change.
- The PR fixes only the lookup typo. Whether the **values** in the `overlay_params` table are correct for every rotation × display combination is a separate question; the table value for rotation=90 on 28c is verified above. If other rotations/displays surface follow-up issues, they can be addressed independently now that the params are actually being applied.

Closes #340.

### DIFF
--- a/adafruit-pitft.py
+++ b/adafruit-pitft.py
@@ -510,8 +510,14 @@ def update_configtxt(rotation_override=None, tinydrm_install=False):
 
     if tinydrm_install: # Use overlay params for Wayland
         overlay += ",drm"
-        if "overlay_params" in pitft_config and pitftrot in pitft_config["overlay_params"] and pitft_config["overlay_params"][pitftrot] is not None:
-            overlay += "," + pitft_config["overlay_params"][pitftrot]
+        # overlay_params lives under the touchscreen sub-dict (the lookup
+        # used to read it from the top level of pitft_config, which never
+        # had the key — see issue #340).
+        touch_cfg = pitft_config.get("touchscreen", {})
+        params_by_rot = touch_cfg.get("overlay_params", {})
+        rot_params = params_by_rot.get(pitftrot)
+        if rot_params:
+            overlay += "," + rot_params
     config_text_base = shell.load_template("templates/config_text_base.txt", date=shell.date(), overlay=overlay)
     if config_text_base is None:
         shell.bail("Unable to load config_text_base template!")


makermelissa (approved):
Looks good.

Issue #253, opened 2023-07-05
rgb-matrix.sh: Update disabling sound for quality mod for RGB matrix HAT


rohithm8:
This script allows selection between two different pinouts corresponding to the quality mod (soldering a wire between gpio 4-18) and the unaltered HAT. However, the quality mod requires disabling onboard sound. Although the method in  `rgb-matrix.sh` worked for older versions of Raspbian, for the latest Raspberry Pi OS I think that it is necessary to blacklist it in the kernel (as commented in the script). Perhaps something like:
`
sudo echo -n "blacklist snd_bcm2835" > /etc/modprobe.d/alsa-blacklist.conf
`
will do the job.

Reference to where this issue is:
https://github.com/adafruit/Raspberry-Pi-Installer-Scripts/blob/339cccfbdd8b503b53186176ff96bead9a13a2f5/rgb-matrix.sh#L276C1-L278C69

Thank you for providing this resource - it has made it significantly easier to use this HAT!

makermelissa:
This has been resolved.

PR #387, opened 2026-06-01
rgb-matrix: add CPU core isolation option for multi-core Pis

makermelissa-ai-assistant:
Adds an optional prompt to reserve a CPU core for driving the LED matrix display by appending `isolcpus=N` to `cmdline.txt`. This is the upstream recommendation from [hzeller/rpi-rgb-led-matrix](https://github.com/hzeller/rpi-rgb-led-matrix) for reducing flicker when the system is busy with other work.

### Behavior

- **Prompt only shown on multi-core Pis** (`$(nproc --all) >= 2`). Skipped on Pi 1 / Zero / Zero W where there is nothing to isolate.
- **N is computed dynamically** as `$(nproc --all) - 1` so it works on both 2-core and 4-core Pis. Pi Zero 2 W / 3 / 4 / 5 all get `isolcpus=3`.
- **Boot path detection:** `CMDLINE_FILE` defaults to `/boot/firmware/cmdline.txt` with a fallback to `/boot/cmdline.txt` for pre-Bookworm releases.
- **Fully idempotent and reversible.** Any existing `isolcpus=<digits>` token in `cmdline.txt` is stripped before the current N is appended, so re-running the installer — or migrating an SD card between Pis with different core counts — does not leave stale or duplicate tokens behind. Selecting *do not reserve* removes the token cleanly.

### Verification

Tested on a Raspberry Pi 4 (4 cores, Trixie) by exercising the `cmdline.txt` edits against a temporary copy:

- Clean add → `isolcpus=3` appended correctly.
- Double add → idempotent (token count stays at 1).
- Synthetic stale `isolcpus=7` token → cleaned up before correct N is appended.
- Round-trip add+remove → file restored byte-exactly to the original.
- Remove when no token present → no-op.

The full interactive prompt flow was also dry-run end-to-end (cancelling before the install step) to confirm the prompt appears in the right order and the summary line displays the chosen option.

### Relationship to #336

This supersedes #336 (originally proposed by @sciguy14) by salvaging its still-novel `isolcpus` feature on top of current `main`. The other improvements from that PR have already landed independently in `main` via subsequent commits:

- Bookworm `/boot/firmware/config.txt` path — done.
- Drop `--force-yes` from `apt-get` invocations — done (Python 2 support was removed entirely).
- `blacklist snd_bcm2835` for quality mode (also addresses #253) — done.
- Newer upstream RGB matrix commit — done (main is at `7a503494…`, newer than #336's bump).

#336 also had a small typo (`CONFIG_FILE=/boot/cmdline.txt` where `CMDLINE_FILE=` was intended) that this rewrite avoids.

cc @sciguy14 — thanks for the original idea.

### DIFF
--- a/rgb-matrix.sh
+++ b/rgb-matrix.sh
@@ -26,6 +26,16 @@ if [ $HAS_PYTHON2 ] && ! [ $HAS_PYTHON3 ]; then
 	exit 0
 fi
 
+NUM_CORES=$(nproc --all)
+# Reserve the highest-numbered core (isolcpus is 0-indexed)
+ISOLCPU_CMD="isolcpus=$((NUM_CORES - 1))"
+
+# Boot config locations. Bookworm and later use /boot/firmware/;
+# pre-Bookworm releases use /boot/.
+CMDLINE_FILE=/boot/firmware/cmdline.txt
+if [ ! -f "$CMDLINE_FILE" ]; then
+	CMDLINE_FILE=/boot/cmdline.txt
+fi
 
 clear
 
@@ -52,6 +62,7 @@ fi
 INTERFACE_TYPE=0
 INSTALL_RTC=0
 QUALITY_MOD=0
+ISOL_CPU=0
 #SLOWDOWN_GPIO=5
 #MATRIX_SIZE=3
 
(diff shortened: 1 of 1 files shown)


makermelissa (approved):
Thank you

Issue #313, opened 2024-05-13
arcade-bonnet.sh install error on raspberry pi 5 with bookworm "externally managed environment"


jareddonovan:
### Script Command

```shell
sudo bash arcade-bonnet.sh
```

### Operating System

Raspberry Pi OS (64-bit) (debian bookworm) with raspberry pi desktop.

### Hardware

Raspberry Pi 5

### Behavior

An error occurs when I run the script:

```
error: externally-managed-environment
```

Otherwise, the script appears to exit as expected and copies the configuration files to the right places. However, when I reboot the Pi, and test the arcade buttons they do not work.

### Description

The full output contains some additional information which may be useful:

```
error: externally-managed-environment

This environment is externally managed

To install Python packages system-wide, try apt install
python3-xyz, where xyz is the package you are trying to
install.

If you wish to install a non-Debian-packaged Python package,
create a virtual environment using python3 -m venv path/to/venv.
Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
sure you have python3-full installed.

For more information visit http://rptl.io/venv

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

I selected following options when the script first runs:

* disable overscan: N
* install GPIO utility: N

Otherwise, it is a fresh install of the operating system (from the raspberry PI imager) with no other added libraries or changes.

### Additional information

I believe the error arises at line 114. When I comment this line out and run the script, it does not give the error (but obviously still does not work):

```
pip3 install evdev smbus
```

I tried several things to fix this. I re-ran the script, rebooted and tested the buttons again after each one, but no luck:

* I deleted the `EXTERNALLY-MANAGED` file in `/usr/lib/python3.11`
* I set up a python virtual environment `python3 -m venv venv`

I have also just tried to edit the `arcade-bonnet.sh` script to use `pipx` instead of `pip`. I will come back and edit the results of this issue with the results of that once I reboot :).

jareddonovan:
Following up on this - after rebooting, following trying `pipx` instead of `pip` in the `arcade-bonnet.sh` script, the buttons still did not work.

I also tried installing the `evdev` and `smbus` libraries with the following commands, as mentioned in the error message:

```
apt install python3-evdev
apt install python3-smbus
```

[...]

PR #388, opened 2026-06-01
joy-bonnet: modernize for Bookworm/Trixie on Pi Zero series

makermelissa-ai-assistant:
## What

Fixes the Joy Bonnet install script for current Raspberry Pi OS releases (Bookworm and Trixie) on Pi Zero, Pi Zero W, and Pi Zero 2 W — the products this script actually targets per the [Joy Bonnet learn guide](https://learn.adafruit.com/adafruit-joy-bonnet-for-raspberry-pi/install-and-use).

## Why

The script has three independent bugs that each prevent it from working on Bookworm+:

1. **PEP 668 / externally-managed-environment.** `pip3 install evdev --upgrade` fails on Bookworm and Trixie because the system Python is now marked externally managed.
2. **`/etc/rc.local` no longer exists.** The autostart step edits `/etc/rc.local`, but Bookworm and Trixie don't ship that file, so the edits are silently dropped and `joyBonnet.py` never starts at boot — even if the user works around bug #1.
3. **`/boot` moved to `/boot/firmware`.** `joyBonnet.py` is installed to `/boot`, but on Bookworm+ the boot partition is mounted at `/boot/firmware` and `/boot` is just an ext4 directory on the rootfs. The systemd unit needs the right path.

This is the same pattern as the issue reported for arcade-bonnet.sh in #313 — pip is the visible symptom; rc.local is the deeper reason buttons still don't work after fixing pip.

## Changes

- Install `python3-evdev` and `python3-smbus` via `apt` instead of pip. Both are packaged in Bookworm and Trixie.
- Drop the `/etc/rc.local` edits; install a `joy-bonnet.service` systemd unit (template matches the pattern already used in `adafruit_fanservice.py` and `pi-eyes.sh`).
- Detect `/boot/firmware` and fall back to `/boot` so the script still works on older Bullseye-and-earlier installs.
- Move the optional GPIO-halt utility to a `gpio-halt.service` systemd unit for the same reason.
- Validate the halt-pin input (BCM 0-27).
- Updated the docstring with target hardware and the rationale for each change.

## Testing

Ran the modernized script end-to-end on Raspberry Pi OS (Trixie, 64-bit) on a Pi 4. Pi 4 is a stand-in for Pi Zero 2 W from an OS-environment perspective (same Bookworm/Trixie codepaths, same apt packages, same `/boot/firmware` layout, same systemd). Verified:

- `apt install python3-evdev python3-smbus` succeeds (no PEP 668 error).
- `/boot/firmware` correctly detected as the boot directory.
- `joyBonnet.py` downloaded and copied into the boot partition.
- `/etc/systemd/system/joy-bonnet.service` written; `systemctl is-enabled joy-bonnet.service` returns `enabled`.
- I2C enabled via `raspi-config nonint do_i2c 0`.
- udev rule written to `/etc/udev/rules.d/10-retrogame.rules`.
- Reboot prompt honors 'N'; script exits 0.

The test artifacts were cleaned off the test rig afterward (no Joy Bonnet hardware is wired to it).

**Not tested:** Actual button presses with physical Joy Bonnet hardware. Happy to hold for a hardware-verification pass if a Joy Bonnet can be brought to the test rig.

## Notes

- The upstream `joyBonnet.py` (in `adafruit/Adafruit-Retrogame`) uses `RPi.GPIO` and `smbus`. Both work natively on Pi Zero, Zero W, and Zero 2 W under current OS releases — no shim required. (`rpi-lgpio` would be required on Pi 5, but Pi 5 is not a target for this product.)
- `arcade-bonnet.sh` has the same three bugs and #313 reports them. I left it out of this PR to keep the change focused; happy to open a sibling PR converting `arcade-bonnet.sh` to Python with the same fixes if that's wanted.

### DIFF
--- a/joy-bonnet.py
+++ b/joy-bonnet.py
@@ -3,26 +3,129 @@ Adafruit Raspberry Pi Joy Bonnet Setup Script
 (C) Adafruit Industries, Creative Commons 3.0 - Attribution Share Alike
 
 Converted to Python by Melissa LeBlanc-Williams for Adafruit Industries
+
+Target hardware: Raspberry Pi Zero, Pi Zero W, Pi Zero 2 W.
+Target OS:      Raspberry Pi OS (Bookworm or newer; 32-bit recommended
+                for full Pi Zero family compatibility).
+
+Notes on modernization (2026):
+  - PEP 668 / Bookworm: Python libraries are installed via apt
+    (python3-evdev, python3-smbus) instead of pip, since system pip
+    is now externally managed.
+  - rc.local is deprecated on Bookworm/Trixie. Autostart is provided
+    via a systemd unit (joy-bonnet.service) instead of editing
+    /etc/rc.local.
+  - The boot partition moved from /boot to /boot/firmware on
+    Bookworm. We detect and prefer /boot/firmware when present.
 """
+
+import os
 
 try:
     from adafruit_shell import Shell
 except ImportError:
-    raise RuntimeError("The library 'adafruit_shell' was not found. To install, try typing: sudo pip3 install adafruit-python-shell")
+    raise RuntimeError(
+        "The library 'adafruit_shell' was not found. To install, try typing: "
+        "sudo pip3 install --break-system-packages adafruit-python-shell"
+    )
 
 shell = Shell()
 shell.group = "JOY"
 
+JOY_BONNET_URL = (
+    "https://raw.githubusercontent.com/adafruit/Adafruit-Retrogame/master/joyBonnet.py"
+)
+GPIO_HALT_URL = (
+    "https://github.com/adafruit/Adafruit-GPIO-Halt/archive/master.zip"
+)
+SERVICE_PATH = "/etc/systemd/system/joy-bonnet.service"
+GPIO_HALT_SERVICE_PATH = "/etc/systemd/system/gpio-halt.service"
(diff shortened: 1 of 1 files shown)


makermelissa-ai-assistant:
Broadened scope: the script now installs cleanly on **all Raspberry Pi models** (Pi Zero / Zero W / Zero 2 W / 3 / 4 / 5).

Added a Pi-5 branch for the Python GPIO library because the legacy `RPi.GPIO` package does not support the RP1 GPIO controller on Pi 5:

- Pi 5 and newer → `python3-rpi-lgpio` (drop-in RPi.GPIO replacement, routes through `lgpio`)
- Pi 4 and earlier → `python3-rpi.gpio` (the real one)

Detection uses `shell.is_pi5_or_newer()`, the same helper already used by `adafruit-pitft.py` and `raspi-blinka.py` in this repo.

The two packages declare a mutual apt `Conflicts:`/`Provides:` relationship, so apt enforces single-version installs and transparently swaps the previous one out if needed.

Re-tested on Pi 4 Trixie: Pi 5 detection correctly returned `False`; apt swapped `python3-rpi-lgpio` (preinstalled on Pi OS) out for `python3-rpi.gpio`; rest of the script ran clean. Pi 5 branch not physically verified (no Pi 5 + Joy Bonnet rig), but uses the same package set as the existing `pi-eyes.sh` Pi 5 branch.

makermelissa (approved):
Thanks, looks good.

PR #382, opened 2026-05-25
Add SPDX license headers to i2smic.py and i2s_mic_module/

makermelissa-ai-assistant:
Closes #243

Adds SPDX license headers to clarify licensing on the files called out in the issue:

- `i2smic.py` → **MIT** (Adafruit-authored installer script, per the convention used for our other installer scripts).
- `i2s_mic_module/snd-i2smic-rpi.c`
- `i2s_mic_module/snd-i2smic-rpi.h`
- `i2s_mic_module/Makefile`
- `i2s_mic_module/dkms.conf`
- `i2s_mic_module/README.md` → all **GPL-2.0-or-later**, matching the existing `MODULE_LICENSE("GPL v2")` declaration in `snd-i2smic-rpi.c` (the kernel maps the "GPL v2" string to GPL-2.0-or-later).

Headers follow the SPDX / REUSE convention already used by `st7789_module/` in this repo (`SPDX-License-Identifier:` on its own line, using `//` for C, `#` for Makefile/dkms.conf/Python, and an HTML comment for the README).

Copyright lines pulled from `git log` and the existing banner comment in `snd-i2smic-rpi.c` (Carter Nelson for Adafruit's Pi4 mods; Huan Truong and Paul Creaser as the original authors).

Follow-up issue #244 covers adding an automatic header check across the repo.

Issue #243, opened 2023-03-21
License Clarification for i2s_mic_module files


jalapenopuzzle:
I request licensing clarification for the following files:

Raspberry-Pi-Installer-Scripts/i2smic.py
Raspberry-Pi-Installer-Scripts/i2s_mic_module/Makefile
Raspberry-Pi-Installer-Scripts/i2s_mic_module/README.md
Raspberry-Pi-Installer-Scripts/i2s_mic_module/dkms.conf
Raspberry-Pi-Installer-Scripts/i2s_mic_module/snd-i2smic-rpi.c
Raspberry-Pi-Installer-Scripts/i2s_mic_module/snd-i2smic-rpi.h

Thank you very much for making these awesome resources available!

jalapenopuzzle:
Oh see that snd-i2smic-rpi.c contains the line
MODULE_LICENSE("GPL v2");

Unfortunately it is not obvious whether this also applies to all the other files listed.

makermelissa:
Good call. I think it's the same for everything in the i2s_mic_module folder. Forthe installer script itself, `i2smic.py` is based on https://github.com/adafruit/Raspberry-Pi-Installer-Scripts/blob/main/converted_shell_scripts/i2smic.sh, which didn't specify a license, so I'm just going to say the MIT license, which is what we pretty much put on all of the stuff we write.

Anyways, having licensing headers on everything (along with an automatic check) is a good idea. I created an issue for adding this.

PR #389, opened 2026-06-05
rtc.py: fix bugs and modernize

makermelissa-ai-assistant:
## What this changes

Fixes a pile of bugs in `rtc.py` that have been there since the original
shell-to-Python conversion. The file's docstring has carried the
`Note: Currently Untested` marker since 2020, and the bugs below are
mostly the kind of thing that would have shown up the first time anyone
actually ran it on hardware.

This PR matches the modernization pattern from joy-bonnet.py (#388) and
the SPDX-header cleanup from #382 / issue #243.

The shell version (`rtc.sh`) is left in place, per the conversion plan
to retire shell scripts to `converted_shell_scripts/` only after the
matching Learn guides are updated.

## Bugs fixed

1. **Arch check excluded every 64-bit Pi.** The original
   `not shell.is_armhf() or shell.is_arm64()` bail meant the script
   refused to run on **any** Raspberry Pi running 64-bit Bookworm or
   Trixie — i.e. the default install on Pi 4 and Pi 5. Replaced with
   `not (shell.is_armhf() or shell.is_arm64())`.

2. **`osreleases = ("Raspbian")` was a bare string, not a tuple.** The
   single-element tuple needs a trailing comma. As written, the `in`
   check on line 87 fell back to substring matching against the OS
   release name. Not a hot path (the script bailed before reaching it
   in most cases) but a real foot-gun.

3. **The "enable I2C?" prompt never fired.** The branch guard was
   `if i2creq == "yes":` but `i2creq` is a Python bool (`True`). Result:
   the RTC overlay was added to config.txt on systems that hadn't been
   configured for I2C yet. The shell script does prompt for this.

4. **SyntaxWarnings on Python 3.12+:**
   - Line 120: `"    \curl -sS $baseurl/$scriptname\n"` — `\c` isn't a
     valid escape sequence, raises `SyntaxWarning: invalid escape
     sequence '\c'` (slated to become a hard `SyntaxError`). Also the
     `$baseurl/$scriptname` was never substituted, so the user just saw
     literal bash-style variables. Removed the whole leftover-from-bash
     line since it never worked.
   - Line 155: the `sed` invocation used `\[` `\/` in a non-raw string,
     producing the same warning. Made the string raw.

5. **`shell.run_command("sudo sed '/.../d' …")` for blanking the
   pre-existing dtoverlay line** — replaced with `shell.pattern_replace`
   using Python regex (`\s`) rather than the POSIX `[[:space:]]` class
   that the shell version relied on. The matching style is what
   `adafruit-pitft.py` / `i2samp.py` / `pi-touch-cam.py` already use
   for config.txt edits. The hwclock-set sed range address (`/pat/,+2`)
   has no clean `pattern_replace` equivalent, so that one still shells
   out — same behavior as before.

6. **No trailing newline at EOF.** Minor, but fixed.

## Cleanup

- Added SPDX license header to match the i2smic.py treatment from #382.
- Moved `shell.require_root()` outside `main()` to match the
  joy-bonnet.py / retrogame.py pattern (per maintainer request). The
  dead `forcesudo` flag goes with it.
- Refactored the "Do you wish to continue?" branch to use an early
  abort return instead of nesting the entire happy path inside an
  `if` block. Smaller indent stack, easier to follow.
- Removed the large pile of dead pimoroni-template flags
  (`debugmode`, `debuguser`, `spacereq`, `armv6`/`7`/`8`, `i2sreq`,
  `spireq`, `uartreq`, `macosxsupport`, `squeezesupport`,
  `jessiesupport`, `armhfonly`, `BOOTCMD`, `APTSRC`, `INITABCONF`,
  `BLACKLIST`, `LOADMOD`, `DTBODIR`, `DEVICE_TREE`, `ASK_TO_REBOOT`,
  `CURRENT_SETTING`, `UPDATED_DB`) that the script never read.
- Dropped the stale `Note: Currently Untested` line from the docstring.

## How I verified

I don't have a physical RTC HAT on hand, but I ran the script end-to-end
on a Raspberry Pi 4 running Trixie (`aarch64`) with the destructive
calls mocked (`apt-get remove`, `update-rc.d`, `sed /lib/udev/...`, the
reboot prompt) and a synthetic `config.txt` pointed at by the
`get_boot_config` mock. The script:

- got past the arch check on `aarch64` — this is the headline bug fix,
  the previous version bailed immediately on this same Pi with `"This
  hardware is not supported, sorry!"`;
- prompted to enable I2C and called `raspi-config do_i2c 0` — the
  previous version silently skipped this entire branch;
- rendered the `select_n` picker, accepted `ds1307`;
- correctly blanked the pre-existing `dtoverlay = i2c-rtc,ds1307`
  line in the synthetic config.txt and appended the new
  `dtoverlay=i2c-rtc,ds1307` at the end;
- ran with the `-y` shortcut to auto-accept the continue prompt.

Running the **pre-PR version** against the same harness reproduces the
two `SyntaxWarning`s and the immediate arch-bail on this `aarch64`
host.

Hardware testing with a real RTC HAT would still be welcome before
this lands — happy to defer to anyone who has a DS1307 / DS3231 /
PCF8523 board lying around. The destructive paths (`apt-get remove
fake-hwclock`, hwclock-set sed) are unchanged from the previous Python
version.

## Out of scope

- `retrogame.py` has a separate set of bugs from its own conversion.
  Those will land in a follow-up PR rather than getting mixed in here.
- The shell version `rtc.sh` stays in place. Move to
  `converted_shell_scripts/` should follow once the Learn guide has
  been updated to reference the Python version.

### DIFF
--- a/rtc.py
+++ b/rtc.py
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: 2020 Melissa LeBlanc-Williams for Adafruit Industries
+#
+# SPDX-License-Identifier: MIT
+
 """
 Adafruit Raspberry Pi Real Time Clock Setup Script
 
@@ -17,9 +21,6 @@ was written for this script.
 Originally written by the - The Pimoroni Crew -
 
 Converted to Python by Melissa LeBlanc-Williams for Adafruit Industries
-
-
-Note: Currently Untested
 """
 
 try:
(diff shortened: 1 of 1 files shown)


makermelissa-ai-assistant:
## Real-hardware verification on Pi 5 + DS1307

Tested today with an Adafruit RGB Matrix HAT + RTC (product 2345) on a
Raspberry Pi 5 running Trixie (Debian 13, kernel 6.12, `aarch64`).

### Setup

Fresh-ish Trixie Lite install. I2C **not** previously enabled, no
pre-existing RTC overlay in `/boot/firmware/config.txt`, `fake-hwclock`
not installed (it's not in the Trixie default package set anymore),
HAT mounted with the standard GPIO4↔GPIO18 mod for matrix PWM (which
doesn't touch any I2C pins so it's irrelevant here).

### Steps and results

```
$ printf "1\n" | sudo python3 rtc.py -y

This script will install everything needed to use Real Time Clock module

RTC --- Warning ---
Always be careful when running scripts and commands
copied from the internet. Ensure they are from a
trusted source.

Checking hardware requirements...

What RTC would you like to install?
[1] ds1307
[2] ds3231
[3] pcf8523

Disabling any current RTC
Adding DT overlay for ds1307
Removing fake-hwclock
RTC Package 'fake-hwclock' is not installed, so not removed
RTC Exiting due to error: Couldn't find /lib/udev/hwclock-set
```

`config.txt` diff after the script ran:

```diff
-#dtparam=i2c_arm=on
+dtparam=i2c_arm=on
...
+dtoverlay=i2c-rtc,ds1307
```

After a reboot:

```
$ ls /dev/i2c-* /dev/rtc*
/dev/i2c-1  /dev/i2c-13  /dev/i2c-14
/dev/rtc -> rtc0   (RP1 internal RTC on Pi 5)
/dev/rtc0          (RP1)
/dev/rtc1          (DS1307 ←  added by the overlay)

$ sudo i2cdetect -y 1
60:                         UU
                            ↑
                  DS1307 at 0x68, claimed by kernel driver

$ sudo dmesg | grep rtc
[1.716] rpi-rtc soc@107c000000:rpi_rtc: registered as rtc0
[1.723] rpi-rtc soc@107c000000:rpi_rtc: setting system clock to 2026-06-05T17:55:53 UTC
[2.853] rtc-ds1307 1-0068: registered as rtc1

$ sudo hwclock -w --rtc=/dev/rtc1   # write system time to DS1307
$ sudo hwclock -r --rtc=/dev/rtc1   # read it back
2026-06-05 10:56:58.009121-07:00    # matches system clock ✅
```

### What the test confirms

Every one of the bug fixes in this PR is exercised end-to-end and
behaves correctly:

- **Arch check no longer bails on aarch64.** Old version would have
  exited immediately with `This hardware is not supported, sorry!` on
  this Pi 5; new version proceeds.
- **I2C-enable branch fires.** The previously-commented `dtparam=i2c_arm=on`
  got uncommented by `raspi-config do_i2c 0`. On the pre-PR version
  this branch was dead because `if i2creq == "yes":` is always false
  for a Python `True`.
- **`select_n` picker works.** Accepted the piped `1` and selected
  `ds1307` cleanly.
- **`pattern_replace` does its job.** No pre-existing RTC line on this
  install so there was nothing to blank, but the call ran without
  error.
- **`dtoverlay=i2c-rtc,ds1307` is appended correctly.** Kernel binds
  the DS1307 to `/dev/rtc1` after reboot, `i2cdetect` shows the driver
  has claimed `0x68` (`UU` not just `68`), and `hwclock` round-trips
  cleanly against the chip.
- **No SyntaxWarnings** at parse or runtime.

### Pre-existing bug **not** addressed by this PR

The script also bails with `Couldn't find /lib/udev/hwclock-set` on
Trixie. This is **not** a regression — both the pre-PR `rtc.py` and
the `rtc.sh` original have the same check, and both fail the same way
on Trixie because `util-linux` no longer ships that file (modern
systemd handles RTC sync directly via the kernel rtc0 binding and
`systemd-timesyncd`).

Crucially: **even though the script exits with an error message, the
config.txt edits are already in place** and the DS1307 binds correctly
on reboot. So the install actually works, but the user is told it
failed. Worth its own follow-up PR — probably:

- detect Trixie (or just "`hwclock-set` is missing") and skip the
  hwclock-set sed step with an informational message rather than
  bailing, since on systemd-managed systems the kernel RTC binding
  is sufficient;
- and have `shell.bail()` return a non-zero exit code so scripted
  callers can detect the failure (right now `bail()` exits 0).

Happy to do that as a follow-up if maintainers agree on the approach.

makermelissa (approved):
Looks much better than before and tested successfully on hardware.

PR #393, opened 2026-06-09
Convert arcade-bonnet.sh to arcade-bonnet.py (fixes #313)

### LINKED ISSUE #313: arcade-bonnet.sh install error on raspberry pi 5 with bookworm "externally managed environment"

makermelissa-ai-assistant:
Closes #313.

Replaces the legacy `arcade-bonnet.sh` shell installer with a Python script (`arcade-bonnet.py`) using `adafruit_shell`, mirroring the `joy-bonnet.py` pattern.

## Why

Issue #313 has two root causes that both broke Arcade Bonnet install on Pi 5 + Bookworm/Trixie:

1. **PEP 668 / `externally-managed-environment`** — `arcade-bonnet.sh` calls `pip3 install evdev smbus`, which Bookworm and newer reject because system Python is externally managed.
2. **Pi 5 GPIO incompatibility** — `arcadeBonnet.py` uses `RPi.GPIO`, which does not support the RP1 GPIO controller on Pi 5 and fails with `RuntimeError: Cannot determine SOC peripheral base address`. The user-reported workaround is to install `python3-rpi-lgpio` (a drop-in replacement that routes the same API through lgpio).

Both are now handled in the installer.

## What changed

- New: `arcade-bonnet.py` using `adafruit_shell` (same shape as `joy-bonnet.py`).
- Moved: `arcade-bonnet.sh` → `converted_shell_scripts/arcade-bonnet.sh` (matches the pattern set by `joy-bonnet.sh`, `i2samp.sh`, etc.).

Behavior:

- Installs deps via apt: `python3-evdev`, `python3-smbus`, and either `python3-rpi.gpio` (Pi 4 and earlier) or `python3-rpi-lgpio` (Pi 5 and newer), gated on `shell.is_pi5_or_newer()`.
- Drops `arcadeBonnet.py` into the active boot partition (`shell.get_boot_config()` → `/boot/firmware` on Bookworm/Trixie, `/boot` on older).
- Installs an `arcade-bonnet.service` systemd unit instead of editing `/etc/rc.local`. Cleans up legacy `arcadeBonnet.py` and `gpio-halt` lines from `/etc/rc.local` if present, so an upgrade from Bullseye doesn't end up double-launching.
- `WorkingDirectory=/tmp` on the systemd unit (rather than the boot dir): on Pi 5 importing `RPi.GPIO` loads lgpio, which creates `.lgd-nfy*` notification pipes in the CWD and `fchmod()`s them. `/boot/firmware` is vfat and rejects permission bits, so the import fails with FileNotFoundError before it completes. `/tmp` is tmpfs and works cleanly. This is a real Pi-5-only gotcha that bites every adafruit_shell installer that pins WorkingDirectory to the boot dir + uses any RPi.GPIO-importing script.
- Optional gpio-halt also moves to a systemd unit (`gpio-halt.service`) instead of an rc.local line.
- udev rule unchanged (`SUBSYSTEM==\"input\", ATTRS{name}==\"retrogame\", ENV{ID_INPUT_KEYBOARD}=\"1\"`).

## Verification

Tested on a Raspberry Pi 5 running Debian Trixie (kernel 6.12) with an Adafruit Arcade Bonnet attached:

- Install script completes with no PEP 668 errors, no `RPi.GPIO` import errors.
- `arcade-bonnet.service` comes up `active (running)`, no journal errors.
- `/dev/input/event5` enumerates as `Name=\"retrogame\"` with `Handlers=sysrq kbd event5` and `ID_INPUT_KEYBOARD=1` from the udev rule.
- `evtest` confirms physical joystick directions on the bonnet translate to `KEY_UP` / `KEY_DOWN` / `KEY_LEFT` / `KEY_RIGHT` events with proper press/release pairs and SYN_REPORT framing — i.e. the full MCP23017 → IRQ → I2C INTCAP read → uinput pipeline works end-to-end.

## Notes

- I2S audio is still NOT installed by this script. `i2samp.py` (already converted) handles that, same as before.
- Closes #313.


makermelissa (approved):
Thank you

Issue #226, opened 2022-07-01
AttributeError: 'NoneType' object has no attribute 'glActiveTexture'


retikulumx:
On a Raspberry Zero, fresh installed, I get the following error

`Traceback (most recent call last):
  File "/boot/Pi_Eyes/eyes.py", line 9, in <module>
    import pi3d
  File "/usr/local/lib/python3.9/dist-packages/pi3d/__init__.py", line 10, in <module>
    from pi3d.constants import (USE_PYGAME, PIL_OK,
  File "/usr/local/lib/python3.9/dist-packages/pi3d/constants/__init__.py", line 198, in <module>
    PLATFORM, bcm, openegl, opengles = _detect_platform_and_load_libraries()
  File "/usr/local/lib/python3.9/dist-packages/pi3d/constants/__init__.py", line 193, in _detect_platform_and_load_libraries
    set_gles_function_args(opengles) #function defined in constants/gl.py
  File "/usr/local/lib/python3.9/dist-packages/pi3d/constants/gl.py", line 387, in set_gles_function_args
    gles.glActiveTexture.argtypes = [GLenum] #
AttributeError: 'NoneType' object has no attribute 'glActiveTexture'`

MNTrailBoss:
Ditto.    Right after reboot from the pieyes load.

PaintYourDragon:
From the forums, somewhat different hardware but same error, unsure of fix but something to try:

> Not sure what to suggest here. Just tested with 2022-09-22-raspios-buster-armhf-lite.img.xz, ran pi-eyes.sh for 240x240 screens, reboot when prompted and it runs successfully on both Pi 3 B and B+.
> 
> Just rampant speculation because I don’t know what else it could be, but if it’s a slower SD card, maybe some prerequisite service is taking some time to load. Try adding a delay (e.g. 10 seconds) in /etc/rc.local before eyes.py is run:
> `cd /boot/Pi_Eyes;sleep 10;python3 eyes.py --radius 240 &`

bikerglen:
[...]

PR #394, opened 2026-06-10
pi-eyes: fix Pi 5 SPI1 right-eye display (fixes #295)

### LINKED ISSUE #295: Adafruit Animated Eyes Bonnet for Raspberry Pi 5 (Bookworm)

makermelissa-ai-assistant:
## Problem

On Raspberry Pi 5, the Adafruit Animated Snake Eyes Bonnet only ever
lights up the **left** eye. The right eye (driven via SPI1) stays
permanently black even though the backlight is on, the panel is
healthy, and the spidev nodes (`/dev/spidev1.0`/`.1`/`.2`) all
appear under `/dev`.

## Root cause

The stock `spi1-3cs` overlay (in `raspberrypi/firmware`) declares
its data-pin pinmux using the legacy `brcm,pins` / `brcm,function`
format from the BCM283x era. The **RP1 pinctrl driver on the Pi 5
silently ignores that format** — there is no warning in dmesg, the
overlay loads, and the spidev nodes appear, but `GPIO19` / `GPIO20`
/ `GPIO21` are never actually muxed to the SPI1 controller.

The SPI1 chip-selects on `GPIO16` / `GPIO17` / `GPIO18` still
work because they're plain GPIOs, so software-side `spidev` writes
appear to succeed and CS pulses correctly, but the panel receives no
usable MOSI/SCLK data — hence the dead-black display.

You can confirm this state with `pinctrl-mux`:

```
# Without the fix
pin 19 (gpio19): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 20 (gpio20): pwr_button     (GPIO UNCLAIMED)
pin 21 (gpio21): (MUX UNCLAIMED) (GPIO UNCLAIMED)

# With the fix
pin 19 (gpio19): 1f00054000.spi function spi1 group gpio19
pin 20 (gpio20): 1f00054000.spi function spi1 group gpio20
pin 21 (gpio21): 1f00054000.spi function spi1 group gpio21
```

## Fix

The Pi 5 base device tree already contains a correct RP1-native pin
group for SPI1:

```
rp1_spi1_gpio19 {
    function = "spi1";
    pins = "gpio19", "gpio20", "gpio21";
    drive-strength = <0x0c>;
    bias-disable;
    slew-rate = <0x01>;
};
```

This PR adds a small overlay (`overlays/spi1-rp1-eyes-overlay.dts`)
that re-points the SPI1 controller's `pinctrl-0` at that node, and
teaches `pi-eyes.sh` to compile and enable it on Pi 5.

Apply order in `/boot/firmware/config.txt` — `spi1-rp1-eyes` MUST
load **after** `spi1-3cs` so its `pinctrl-0` assignment wins:

```
dtparam=spi1=on
dtoverlay=spi1-3cs
dtoverlay=spi1-rp1-eyes
```

## Changes

- **New:** `overlays/spi1-rp1-eyes-overlay.dts` — 8-line overlay,
  compiles to a ~360-byte `.dtbo` with a single phandle fixup for
  `rp1_spi1_gpio19`. Safe to load on Pi 3B/Pi 4 too: the fixup
  symbol only exists in the RP1/Pi-5 DTB, so `dtoverlay` simply
  skips the apply on older Pis where `spi1-3cs` already works.
- **`pi-eyes.sh`:**
  - Installs `device-tree-compiler` (for `dtc`).
  - On Pi 5, compiles the new overlay to
    `/boot/firmware/overlays/spi1-rp1-eyes.dtbo` and appends
    `dtoverlay=spi1-rp1-eyes` to `config.txt` after `spi1-3cs`.
  - Tightens the existing `spi1-3cs` reconfig regex from
    `^.*dtoverlay=spi1.*$` to `^.*dtoverlay=spi1-3cs.*$` so it
    doesn't clobber the new `dtoverlay=spi1-rp1-eyes` line on
    subsequent runs.

## Testing

Verified on Raspberry Pi 5 + Raspberry Pi OS Trixie + Adafruit
Snake Eyes Bonnet with two 1.54" ST7789 240×240 IPS displays:

- Before the overlay: right eye permanently black, left eye works
  normally. `pinctrl-mux` shows GPIO19/20/21 as
  `(MUX UNCLAIMED)`.
- After the overlay: both eyes light on a solid-color SPI fill test
  via `/dev/spidev0.0` (left) and `/dev/spidev1.2` (right).
  `pinctrl-mux` shows all three pins claimed by
  `1f00054000.spi function spi1`.
- End-to-end with the installed `pi-eyes.sh`, `xinit eyes.py` +
  `fbx2 -i` services run cleanly and both eyes animate.

Fixes #295 (and supersedes the older Pi-5 Pi-Eyes thread at #226,
which has the same root cause).

🤖 Generated with [OpenClaw](https://openclaw.ai) (Piclaw + Claude
Opus 4)

### DIFF
--- a/overlays/spi1-rp1-eyes-overlay.dts
+++ b/overlays/spi1-rp1-eyes-overlay.dts
@@ -1,0 +1,55 @@
+/dts-v1/;
+/plugin/;
+
+/*
+ * spi1-rp1-eyes: fix SPI1 data-pin muxing on Raspberry Pi 5 (RP1)
+ * for the Adafruit Animated Snake Eyes Bonnet.
+ *
+ * Background
+ * ----------
+ * On Pi 3B / Pi 4 (BCM283x), the stock `spi1-3cs` overlay correctly
+ * mux GPIO19 / GPIO20 / GPIO21 to the SPI1 controller as MISO / MOSI
+ * / SCLK using the legacy `brcm,pins` / `brcm,function` pinctrl
+ * format.
+ *
+ * On Pi 5 (RP1), that legacy pinctrl format is silently ignored by
+ * the RP1 pinctrl driver. The result is that `spi1-3cs` brings up
+ * the SPI1 chip-selects (CE0/CE1/CE2 on GPIO18/17/16, which are
+ * plain GPIOs) but never muxes the MOSI / SCLK data pins, so the
+ * right-eye TFT receives a clocking CS but no usable data and the
+ * panel stays permanently black.
+ *
+ * Fix
+ * ---
+ * The Pi 5 base device tree already contains a correct RP1-native
+ * pinctrl node for SPI1 (`rp1_spi1_gpio19`, function = "spi1",
+ * pins = "gpio19", "gpio20", "gpio21"). This overlay simply
+ * re-points the SPI1 controller's `pinctrl-0` at that node,
+ * replacing `spi1-3cs`'s broken legacy entry.
+ *
+ * Apply order in /boot/firmware/config.txt — this overlay MUST be
+ * loaded AFTER `spi1-3cs`, so that its `pinctrl-0` assignment wins:
+ *
+ *     dtparam=spi1=on
+ *     dtoverlay=spi1-3cs
+ *     dtoverlay=spi1-rp1-eyes
+ *
+ * Safe to load on Pi 3B / Pi 4 as well: the `rp1_spi1_gpio19` label
+ * is only present in the RP1/Pi-5 DTB, so dtoverlay will simply
+ * skip this overlay's fixup on older Pis where the legacy
+ * `spi1-3cs` pinctrl already works correctly.
+ *
+ * Issue: https://github.com/adafruit/Raspberry-Pi-Installer-Scripts/issues/295
+ */
+
+/ {
+	compatible = "brcm,bcm2712";
+
+	fragment@0 {
+		target = <&spi1>;
+		__overlay__ {
+			pinctrl-names = "default";
+			pinctrl-0 = <&rp1_spi1_gpio19>;
+		};
+	};
+};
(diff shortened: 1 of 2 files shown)


makermelissa-ai-assistant:
Holding this in draft pending one more empirical check. The device-tree analysis (legacy `brcm,pins` format in `spi1-3cs` being ignored by the RP1 pinctrl driver) is suggestive, and `pinctrl-mux` clearly went from UNCLAIMED to `function spi1 group gpio19` after the overlay loaded on the Pi 5 I was working with. But that same Pi 5 turned out to have an independent hardware fault, and the second (healthy) Pi 5 was only ever tested with the overlay already in place.

Want to confirm on the healthy Pi 5 that removing `dtoverlay=spi1-rp1-eyes` (and rebooting) actually does reproduce a black right eye on stock `spi1-3cs` alone, before asking for review. Will flip to ready once that's verified or close this out if the overlay turns out to be unnecessary.

makermelissa:
This isn't necessary. My Pi 5's SPI1 seems to be broken when verifying that this is fixed causing testing to be incorrect. Swapping to another Pi 5 fixed the issue.

PR #396, opened 2026-06-16
Add uninstall option to retrogame

### LINKED ISSUE #60: Add uninstall option to retrogame
There is currently no option to uninstall retrogame in the installer. Please add one. Thanks.

makermelissa-ai-assistant:
Adds an **Uninstall retrogame** entry to the configuration menu, as requested in #60.

When selected, it:
- Stops and disables `retrogame.service`, removes the unit, and runs `daemon-reload`
- Removes the `/usr/local/bin/retrogame` binary
- Removes the `/etc/udev/rules.d/10-retrogame.rules` udev rule
- Cleans up any legacy `rc.local` autostart line from older installs
- Prompts before removing `/boot/retrogame.cfg` (kept by default, since it's user-editable)

The routine is idempotent — it tolerates any already-absent files so a partial/repeat uninstall won't error out. Ends with the usual reboot prompt.

Closes #60

### DIFF
--- a/retrogame.py
+++ b/retrogame.py
@@ -57,6 +57,49 @@ WantedBy=multi-user.target
     shell.run_command("systemctl enable retrogame.service")
 
 
+def uninstall():
+    """Remove retrogame and all files this installer creates."""
+    print("\nThis will remove retrogame, its systemd service,\nand the udev rule.\n")
+    if not shell.prompt("Uninstall retrogame?", default="n"):
+        print("Canceled.")
+        shell.exit()
+
+    print("Stopping and removing retrogame service...", end="")
+    # Tolerate an already-stopped/absent unit so uninstall stays idempotent.
+    shell.run_command("systemctl stop retrogame.service", suppress_message=True)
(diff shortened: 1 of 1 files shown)


makermelissa-ai-assistant:
Hardware-tested on a Raspberry Pi 4 (Debian Trixie):

- **Install** one config, then **uninstall** via the new menu entry → binary, udev rule, and systemd unit removed; the enable symlink in `multi-user.target.wants/` is gone and `systemctl is-enabled retrogame.service` reports `not-found`.
- **Keep-cfg path** (default): `/boot/retrogame.cfg` correctly preserved.
- **Remove-cfg path**: `/boot/retrogame.cfg` correctly removed when the prompt is answered yes.
- **Idempotent**: running the uninstall again with nothing installed completes cleanly (exit 0, no errors), and skips the cfg prompt since the file is already gone.

`ruff format` + `ruff check` both pass.

makermelissa (approved):
Looks good. Thank you.

